PR #90, opened 2026-02-14
Fix 6 security findings from internal review

luckyPipewrench:
## Summary
- 6 security findings fixed from internal security review
- goconst lint fix for CI v2.9.0 compatibility

## Fixes
- MCP `ExtractText` joins content blocks with space (was `\n`), catching between-word injection splits
- Git `CompileDLPPatterns` adds `(?i)` prefix for case-insensitive matching
- Rate limiter uses `baseDomain()` normalization, preventing per-subdomain evasion
- Response scanning normalizes Ogham/Mongolian whitespace before regex matching
- `ValidateAgentName` rejects names containing `..` or equal to `.`
- URL DLP applies NFKC normalization before pattern matching

## Test plan
- [x] All tests pass with `-race -count=1`
- [x] `golangci-lint v2.9.0 run ./...` reports 0 issues
- [ ] CI passes (test, lint, build)

## Summary by CodeRabbit

* **Bug Fixes**
  * DLP pattern matching is now case-insensitive to prevent evasion.
  * Rate limiting now normalizes domain names to stop subdomain rotation bypasses.
  * Content block concatenation updated to preserve between-block injections.
  * Unicode whitespace normalization added to prevent whitespace-based bypasses.
  * Agent name validation strengthened to block path-traversal names.
  * Applied NFKC Unicode normalization before DLP matching for consistency.


luckyPipewrench:
@coderabbitai review

PR #102, opened 2026-02-16
Enforce strict-mode API allowlist, fix 3 static analysis bugs

luckyPipewrench:
## Summary
- Enforce API allowlist at scanner runtime in strict mode — previously validated but never checked during the 9-layer scan pipeline. Runs after scheme check, before DNS resolution.
- Add `ViolationPermissions` case to integrity text output (was silently dropped in non-JSON mode).
- Warn when config hot-reload switches to `ask` mode but HITL approver was not created at startup. Runtime still blocks (fail-closed), now with explicit diagnostic.
- Fix misleading comment about strict mode proxy behavior.
- Add `ActionAsk` constant, renumber scanner pipeline step comments.

## Test plan
- [x] 8 new allowlist unit tests (block, permit, wildcard, empty, balanced bypass, audit bypass, case insensitive, ordering)
- [x] 1 new ViolationPermissions text output test
- [x] All 1,500+ existing tests pass with `-race`
- [x] 0 lint issues (`golangci-lint run ./...`)
- [x] Pen tested on fedora (90/90 pass including 7 new allowlist tests)

## Summary by CodeRabbit

* **New Features**
  * Added domain allowlist support with strict mode enforcement
  * Added permissions violation reporting in integrity checks
  * Added conditional human-in-the-loop approver initialization

* **Bug Fixes**
  * Added warnings when configuration changes to require manual approval without proper setup

* **Tests**
  * Added comprehensive allowlist validation tests including wildcard matching and case-insensitive behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,8 @@ git_protection:
 ## Integration Guides
 
 - **[Claude Code](docs/guides/claude-code.md)** — MCP proxy setup, `.claude.json` configuration, HTTP fetch proxy hooks
+- **[CrewAI](docs/guides/crewai.md)** — `MCPServerStdio` wrapping, `MCPServerAdapter`, Docker Compose
+- **[LangGraph](docs/guides/langgraph.md)** — `MultiServerMCPClient`, `StateGraph`, Docker deployment
 - Cursor — use `configs/cursor.yaml` with the same MCP proxy pattern as [Claude Code](docs/guides/claude-code.md)
 
 ## CI/CD Usage
--- a/docs/guides/crewai.md
+++ b/docs/guides/crewai.md
@@ -1,0 +1,336 @@
+# Using Pipelock with CrewAI
+
+Pipelock wraps CrewAI's MCP servers as a stdio proxy, scanning every request and
+response for credential leaks, prompt injection, and tool description poisoning.
+This guide covers both the `mcps=` DSL and the `MCPServerAdapter` approach.
+
+## Quick Start
+
+```bash
+# 1. Install pipelock
+go install github.com/luckyPipewrench/pipelock/cmd/pipelock@latest
+
+# 2. Generate a config (or copy a preset)
+pipelock generate config --preset generic-agent > pipelock.yaml
+
+# 3. Verify
+pipelock version
+```
+
+```python
+from crewai import Agent
+from crewai.mcp import MCPServerStdio
+
+agent = Agent(
+    role="Research Analyst",
+    goal="Research and analyze information",
+    backstory="Expert researcher with access to external tools",
+    mcps=[
+        MCPServerStdio(
+            command="pipelock",
+            args=[
+                "mcp", "proxy",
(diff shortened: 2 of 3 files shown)


internal/scanner/scanner.go
```diff
@@ -189,12 +198,18 @@ func (s *Scanner) Scan(rawURL string) Result {
...
 	}
 
-	// 2. Blocklist check — before DNS to avoid resolving known-bad domains.
+	// 2. Allowlist check — if configured, only allowlisted domains are permitted.
+	// Runs before DNS to reject disallowed domains without any network I/O.
+	if result := s.checkAllowlist(hostname); !result.Allowed {
+		return result
+	}
+
+	// 3. Blocklist check — before DNS to avoid resolving known-bad domains.
```
DeathCamel58, line 207: Is this comment really necessary? Asking because by adding a number to this comment bumps all the numbers below, causing additional changes. Not really a problem, just would be a bit cleaner with a smaller diff.

internal/cli/generate.go
```diff
@@ -87,7 +87,7 @@ Examples:
 func strictPreset() *config.Config {
 	cfg := config.Defaults()
 	cfg.Mode = config.ModeStrict
-	// In strict mode, the fetch proxy is not used — agent can only reach allowlisted APIs
+	// In strict mode, the fetch proxy enforces the API allowlist
```
DeathCamel58, line 90: Is this comment change necessary?

internal/cli/run.go
```diff
@@ -79,7 +79,8 @@ Examples:
 			m := metrics.New()
 
 			var proxyOpts []proxy.Option
-			if cfg.ResponseScanning.Action == "ask" {
+			hasApprover := cfg.ResponseScanning.Action == config.ActionAsk
+			if hasApprover {
```
DeathCamel58, line 83: Up to you, but `if cfg.ResponseScanning.Action == config.ActionAsk {` would do the same thing, and wouldn't introduce a new variable. Less changes = better typically

internal/scanner/scanner.go
luckyPipewrench: Fair point on the diff noise. I keep them numbered because the scanner pipeline is 9 layers deep and the numbers make it easier to cross-reference with the docs. But yeah, the renumbering ripple is annoying.

internal/cli/generate.go
luckyPipewrench: Yeah, the old comment said "the fetch proxy is not used" in strict mode which was just wrong. The proxy runs in all modes, and now it actually enforces the allowlist, so the comment needed to match reality.

internal/cli/run.go
luckyPipewrench: The variable gets reused about 50 lines down in the reload handler to check whether we can warn about missing HITL. That's why it's pulled out instead of inlined.

PR #112, opened 2026-02-19
feat(mcp): add Streamable HTTP transport for MCP proxy

luckyPipewrench:
## Summary

- Adds MCP Streamable HTTP transport so `pipelock mcp proxy --upstream <url>` bridges stdio-to-HTTP with full bidirectional scanning
- New SSE reader for server-initiated message streams, HTTP client with JSON/SSE response handling, and integration into the existing MCP proxy pipeline
- All request/response scanning (DLP, prompt injection, tool poisoning, tool policy) applies identically to HTTP transport
- 6-stage shell normalization pipeline for tool policy matching: Unicode → octal/hex decode → backslash strip → command substitution resolve → variable assignment resolve → IFS expansion
- Expanded Instruction Downgrade detection across all config presets
- Adds `pipelock test` command for config validation against attack corpus

## New files

| File | Purpose |
|------|---------|
| `internal/mcp/httpclient.go` | HTTP client with JSON and SSE response handling |
| `internal/mcp/sse.go` | SSE stream reader implementing MessageReader interface |
| `internal/mcp/proxy_http.go` | Stdio-to-HTTP bridge proxy |
| `internal/cli/test.go` | Config validation command with built-in attack corpus |
| `docs/guides/mcp-http-transport.md` | Integration guide |

## Security hardening (from review rounds)

- Shell normalization: octal/hex escape decode, variable concat resolve (`x=rm;$x`), command substitution resolve (`$(printf rm)`)
- Context propagation to HTTP requests with backoff and URL validation
- Instruction Downgrade regex expansion: 5 verb × 7 temporal × 11 noun × 2 connector × 9 adjective coverage
- Fail-closed on HTTP transport errors (timeouts, malformed responses, SSE parse failures)

## Test plan

- [x] `go test -race -count=1 ./...` — all tests pass
- [x] `golangci-lint run ./...` — zero issues
- [x] 232/232 pen test cases pass
- [x] Private white-box Go tests pass (211 functions across 8 packages)
- [x] New unit tests for httpclient, SSE reader, HTTP proxy, policy normalization, CLI test command
- [ ] CI passes (lint, test matrix Go 1.24+1.25, build)
- [ ] CodeRabbit review

## Summary by CodeRabbit

* **New Features**
  * Remote Streamable HTTP transport (--upstream) for the MCP proxy with bidirectional scanning, session lifecycle tracking, and automatic session deletion
  * SSE GET stream support for server-initiated messages and notifications

* **Bug Fixes / Improvements**
  * Stronger scanning: encoded/obfuscated command detection, tool-call safeguards, split-field secret detection, deterministic JSON traversal, and improved normalization/homoglyph handling
  * clearer CLI validation and error messages for proxy usage

* **Documentation**
  * Updated docs and examples for local vs remote MCP proxy

* **Tests**
  * Expanded tests for HTTP proxy, SSE, scanning, and policy rules

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -633,7 +633,7 @@ func Defaults() *Config {
 				{Name: "Authority Escalation", Regex: `(?i)you\s+(now\s+)?have\s+(full\s+)?(admin|root|system|superuser|elevated)\s+(access|privileges|permissions|rights)`},
 				{Name: "Instruction Downgrade", Regex: `(?i)(treat|consider|regard|reinterpret|downgrade)\s+((?:the|all)\s+)?(previous|prior|above|system|policy|original|existing)\s+(\w+\s+)?(text|instructions?|rules|directives|guidelines|safeguards|constraints|controls|checks|context|prompt|policies|guardrails|parameters)\s+((as|to)\s+)?(historical|outdated|deprecated|optional|background|secondary|non-binding|informational|advisory)`},
 				{Name: "Instruction Dismissal", Regex: `(?i)(set|put)\s+(the\s+)?(previous|prior|above|system|original)\s+(\w+\s+)?(instructions?|directives|rules|constraints|context|prompt|safeguards|guidelines|policies|guardrails)\s+(aside|away|to\s+(one|the)\s+side)`},
-				{Name: "Priority Override", Regex: `(?i)prioritize\s+(the\s+)?(task|user|current|new|latest)\s+(request|message|input|instructions?|prompt)`},
+				{Name: "Priority Override", Regex: `(?i)\bprioritize\s+(the\s+)?(task|user|current|new|latest)\s+(request|message|input|instructions?|prompt)`},
 			},
 		},
 		Logging: LoggingConfig{


internal/config/config.go
```diff
@@ -588,16 +588,19 @@ func Defaults() *Config {
...
 				{Name: "Tool Invocation", Regex: `(?i)you\s+must\s+(immediately\s+)?(call|execute|run|invoke)\s+(the|this)\s+(function|tool|command|api|endpoint)`},
 				{Name: "Authority Escalation", Regex: `(?i)you\s+(now\s+)?have\s+(full\s+)?(admin|root|system|superuser|elevated)\s+(access|privileges|permissions|rights)`},
+				{Name: "Instruction Downgrade", Regex: `(?i)(treat|consider|regard|reinterpret|downgrade)\s+((?:the|all)\s+)?(previous|prior|above|system|policy|original|existing)\s+(\w+\s+)?(text|instructions?|rules|directives|guidelines|safeguards|constraints|controls|checks|context|prompt|policies|guardrails|parameters)\s+((as|to)\s+)?(historical|outdated|deprecated|optional|background|secondary|non-binding|informational|advisory)`},
+				{Name: "Instruction Dismissal", Regex: `(?i)(set|put)\s+(the\s+)?(previous|prior|above|system|original)\s+(\w+\s+)?(instructions?|directives|rules|constraints|context|prompt|safeguards|guidelines|policies|guardrails)\s+(aside|away|to\s+(one|the)\s+side)`},
+				{Name: "Priority Override", Regex: `(?i)prioritize\s+(the\s+)?(task|user|current|new|latest)\s+(request|message|input|instructions?|prompt)`},
```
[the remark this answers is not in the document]
luckyPipewrench: This is intentional — the presets are meant to be broader than `Defaults()`. Each preset targets a specific use case (balanced, strict, audit) and includes patterns beyond the base defaults. The instruction dismissal regex here is a superset of what `Defaults()` provides.

internal/mcp/httpclient.go
```diff
@@ -0,0 +1,284 @@
...
+	if r.done {
+		return nil, io.EOF
+	}
+	r.done = true
+
+	data, err := io.ReadAll(io.LimitReader(r.body, int64(maxLineSize)))
+	_ = r.body.Close() // best-effort cleanup after read
+	if err != nil {
+		return nil, fmt.Errorf("reading response body: %w", err)
+	}
```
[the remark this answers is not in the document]
luckyPipewrench: Good catch on the naming, but `maxLineSize` (1 MB) works well here as a general upper bound for a single JSON-RPC message. It prevents unbounded reads while staying large enough for any real MCP response. Adding a separate constant would just be the same value with a different name.

internal/mcp/proxy_http_test.go
```diff
@@ -0,0 +1,1218 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			got := extractRPCID([]byte(tt.msg))
+			if tt.want == "" {
+				if got != nil {
+					t.Errorf("expected nil, got %s", string(got))
+				}
+			} else {
+				if string(got) != tt.want {
+					t.Errorf("got %s, want %s", string(got), tt.want)
+				}
```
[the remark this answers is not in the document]
luckyPipewrench: This is a deliberate design choice. `"id": null` is ambiguous in JSON-RPC 2.0 — it could be a malformed request or a response to something we can't track. We treat it as a notification (no response written) rather than forwarding a message we can't correlate. Fail-closed is the right call here, and the test asserts that behavior intentionally.

internal/mcp/httpclient.go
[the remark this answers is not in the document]
luckyPipewrench: You're right, I misread the suggestion. The real issue is silent truncation, not the constant name. Fixed: now reads `maxLineSize + 1` bytes and returns a clear `"response body exceeds maximum size"` error if it gets truncated. Consistent with how `transport.go:71` and `proxy.go:40` handle it.

internal/mcp/httpclient.go
```diff
@@ -0,0 +1,292 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("sending request: %w", err)
+	}
+
+	// Track session ID from response.
+	if sid := resp.Header.Get("Mcp-Session-Id"); sid != "" {
+		c.sessionMu.Lock()
+		c.sessionID = sid
+		c.sessionMu.Unlock()
+	}
```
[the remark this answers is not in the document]
luckyPipewrench: Good catch. Moved the session ID update inside a `resp.StatusCode < 300` guard so error responses can't overwrite a valid session. Added a test that confirms a 500 with a crafted `Mcp-Session-Id` header doesn't corrupt the established session.

internal/mcp/policy.go
```diff
@@ -265,19 +357,72 @@ func stricterAction(a, b string) string {
...
-		a = "block"
-		ra = actionRank["block"]
+		a = "block" //nolint:goconst // action string used as-is from config
+		ra = actionRank[a]
 	}
 	if !bOK {
-		b = "block"
-		rb = actionRank["block"]
+		b = "block" //nolint:goconst // action string used as-is from config
+		rb = actionRank[b]
```
DeathCamel58: This is functionally the same as the code before this PR. Any reason to modify it?

.claude
DeathCamel58, line 1: I don't think you meant to commit this.

tests
DeathCamel58, line 1: I don't think you meant to commit this.

DeathCamel58 (dismissed):

tests
luckyPipewrench: oops

.claude
luckyPipewrench: nope

internal/mcp/policy.go
luckyPipewrench: no, fixed.

PR #142, opened 2026-02-25
feat: add finding suppression, kill switch, MCP refactor, and chain detection

luckyPipewrench:
## Summary
- **Finding suppression** — inline `pipelock:ignore` comments and config-based `suppress` entries filter findings in static analysis (audit, git scan-diff) and live proxy response scanning. URL suffix matching supports paths like `robots.txt`.
- **MCP package refactor** — split `internal/mcp` into `transport`, `tools`, `policy`, `jsonrpc` sub-packages for maintainability. Zero behavior change.
- **Kill switch** — OR-based activation via config, SIGUSR1 signal, or sentinel file. Blocks all proxy traffic (fetch, forward, WS, MCP) when any source is active. Hot-reload aware.
- **Chain detection** — subsequence matching on classified tool call sequences. 8 built-in attack patterns (read-then-exec, env-then-network, etc.), configurable window/gap/overrides, Prometheus metrics. Session-keyed by `Mcp-Session-Id` with IP-based fallback.

## Test plan
- [x] `golangci-lint run ./...` — 0 issues
- [x] `go test -race -count=1 ./...` — all pass
- [x] Pen test (350 cases) — all pass
- [x] Chain detection end-to-end verified with mock MCP server (read-then-exec, env-then-network, directory-scan patterns confirmed with Prometheus counters)

## Summary by CodeRabbit

* **New Features**
  * Emergency kill‑switch (config/signal/sentinel) with IP allowlist and health/metrics exemptions; runtime toggle and JSON-RPC error responses.
  * Suppression system: inline comments and config-based suppression with verbose suppressed reporting.
  * Session‑aware tool‑chain detection with warn/block actions and chain metrics.
  * Structured JSON‑RPC handling for improved MCP scanning.

* **Enhancements**
  * CLI flags to load suppression config and show suppressed findings; proxies and MCP flows honor kill‑switch, suppression, and chain detections.
  * Metrics for kill‑switch denials and chain detections.

* **Refactor**
  * Reorganized internal packages to expose transport, tools, jsonrpc abstractions.

* **Tests**
  * Extensive unit and integration tests covering new features.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -868,6 +868,13 @@ func (c *Config) Validate() error {
 		if s.Path == "" {
 			return fmt.Errorf("suppress entry %d (%s) missing required field \"path\"", i, s.Rule)
 		}
+		// Validate glob syntax so misconfigured patterns fail fast
+		// instead of silently never matching at runtime.
+		if strings.ContainsAny(s.Path, "*?[") {
+			if _, err := path.Match(toSlash(s.Path), "x"); err != nil {
+				return fmt.Errorf("suppress entry %d (%s) has invalid path pattern %q: %w", i, s.Rule, s.Path, err)
+			}
+		}
 	}
 
 	// Validate kill switch allowlist CIDRs are parseable
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -2494,6 +2494,30 @@ func TestValidate_SuppressNilList(t *testing.T) {
 	}
 }
 
+func TestValidate_SuppressInvalidGlob(t *testing.T) {
+	cfg := Defaults()
+	cfg.Suppress = []SuppressEntry{
+		{Rule: "Credential in URL", Path: "foo[", Reason: "bad glob"},
+	}
+	err := cfg.Validate()
+	if err == nil {
+		t.Fatal("expected error for malformed glob pattern")
+	}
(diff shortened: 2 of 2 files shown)


internal/mcp/scan.go
```diff
@@ -8,149 +8,70 @@ import (
...
-		}
-	}
-	var parsed interface{}
-	if err := json.Unmarshal(raw, &parsed); err == nil {
-		extract(parsed)
-	}
-	return result
-}
+	NewChainMatcher = chains.NewMatcher
+)
```
DeathCamel58: Do we actually need to re-export these, or can we directly reference the definitions?

luckyPipewrench: Resolved.

internal/cli/test.go
```diff
@@ -11,6 +11,7 @@ import (
 
 	"github.com/luckyPipewrench/pipelock/internal/config"
 	"github.com/luckyPipewrench/pipelock/internal/mcp"
+	mcptools "github.com/luckyPipewrench/pipelock/internal/mcp/tools"
```
DeathCamel58, line 14: Do we need this alias?

internal/mcp/scan.go
luckyPipewrench: tools alone would compile fine (there's no collision in this file), but mcptools is clearer about which tools package this is.

Issue #276, opened 2026-03-22
Configuring exceptions to internal networks


Jorilx:
**Use case**
Sometimes an agent must connect to localhost to test the application it's building, or to a local inference server, but it gets blocked by the SSRF protection check.

**Proposed solution**
It would be nice to have a way to configure exceptions to the "internal" networks setting, to allow the agent to connect to the specified hosts.

**Alternatives considered**
The same result could be achieved by modifying the default "internal" setting, for example to allow connecting to 192.168.3.x:
```internal:
  - "0.0.0.0/8"
  - "127.0.0.0/8"
  - "10.0.0.0/8"
  - "100.64.0.0/10"
  - "172.16.0.0/12"
  - "192.168.0.0/24"
  - "192.168.1.0/24"
  - "192.168.2.0/24"
  - "192.168.4.0/24"
  - ...
  - "192.168.254.0/24"
  - "169.254.0.0/16"
  - "::1/128"
  - "fc00::/7"
  - "fe80::/10"
  - "224.0.0.0/4"
  - "ff00::/8"
```
...but obviously it's very verbose

Something like "internal-allowlist" would permit easier configuration:
```
internal-allowlist:
   - 192.168.3.200
   - 127.0.0.1
```

PR #297, opened 2026-03-25
Add support for trusted_domains to forward proxy mode

Jorilx:
## Summary

The trusted_domains settings is checked when in fetch proxy mode, but not while in forward proxy mode, see https://github.com/luckyPipewrench/pipelock/issues/276#issuecomment-4121351417

## Changes

Created a IsTrustedDomain() function, called inside ssrfSafeDialContext()

## Testing

- [x] `make test` passes
- [x] `make vet` passes
- [x] New tests added for new functionality
- [x] Tested manually (see comment on issue 276)

## Security Considerations

It can lower the scanning security, but it's entirely opt-in and configurable

## Summary by CodeRabbit

* **Bug Fixes**
  * SSRF protection now honors trusted-domain allowlists for hostname destinations while still blocking raw IP literals.

* **Tests**
  * Added tests verifying trusted-domain bypass behavior for hostnames.
  * Added tests for trusted-domain matching (wildcards, case/format normalization, and edge cases).


internal/scanner/scanner.go
```diff
@@ -352,6 +352,18 @@ func (s *Scanner) IsInternalIP(ip net.IP) bool {
 	return false
 }
 
+// IsTrustedDomain checks if a hostname matches any trusted domain pattern.
+// Trusted domains bypass the SSRF internal-IP check.
+func (s *Scanner) IsTrustedDomain(hostname string) bool {
+	hostname = strings.ToLower(strings.TrimSuffix(hostname, "."))
```
DeathCamel58: The `MatchDomain` function's first argument is already passed through `strings.ToLower`. There's no need to `ToLower` here.

internal/proxy/proxy.go
```diff
@@ -642,7 +645,9 @@ func (p *Proxy) ssrfSafeDialContext(ctx context.Context, network, addr string) (
 			ip = v4
 		}
 		if currentSc.IsInternalIP(ip) {
-			return nil, fmt.Errorf("SSRF blocked: %s resolves to internal IP %s", host, ipStr)
+			if !currentSc.IsTrustedDomain(host) {
```
DeathCamel58: Minor nitpick: IsTrustedDomain(host) is inside the IP loop, but the host doesn't change inside of the loop. Moving the check before the loop could avoid re-checking every resolved IP.

DeathCamel58 (approved):

PR #392, opened 2026-04-11
feat: per-pattern warn mode for DLP rollout safety

luckyPipewrench:
## Summary

- Operators can annotate DLP patterns with `action: warn` to observe matches in production without blocking traffic. Warn-mode patterns run through the full scanning pipeline (URL DLP, text DLP, encoded variants, cross-request fragment detection) but produce informational results instead of enforcement actions.
- The warn filter is handled at the scanner output layer: text DLP partitions into `Matches` (enforced) and `InformationalMatches` (warn-only); URL DLP collects warn matches in `Result.WarnMatches`. Transports check only `Matches`/`Allowed` and require no changes.
- Built-in default patterns reject warn mode at config validation — the immutable safety floor always enforces.
- A `DLPWarnHook` callback on the scanner package enables audit event emission (`dlp_warn`) from within `ScanTextForDLP` and `Scan` automatically, without per-transport wiring. Runtime hook assignment is deferred to a follow-up once `run.go` is available (#384).

## Design notes

**Scanner-output-layer routing** was chosen over threading warn checks through transports because 5 of 8 transport files are reserved by open PRs (#384, #385, #388, #389). The scanner partitions enforced vs. informational; transports see only enforced matches and need zero changes.

**Named return + defer in `scan()`** propagates warn matches through 7 return paths after the DLP check (entropy, SSRF, rate limit, URL length, data budget, context, final pass) without touching each return statement.

**Audit emission scope:** `DLPWarnHook` is called from within the scanner whenever warn matches are produced. The runtime wires it to `audit.Logger.LogDLPWarn` — that wiring is out of scope for this PR because `internal/cli/runtime/run.go` is reserved by #384. The hook mechanism is tested and ready; wiring is a one-liner once #384 lands.

## Test plan

New test functions cover all warn-mode paths:
- 6-state config validation (omitted, null, warn, non-warn rejected, built-in rejected, reload)
- Text DLP routing (warn → InformationalMatches, enforce → Matches, mixed, encoded variants)
- URL DLP routing (warn allows, enforce blocks, mixed, subsequence)
- Fragment buffer cross-request warn propagation
- DLPWarnHook callback for text and URL paths, nil safety
- Audit event structure and emitter integration

## Summary by CodeRabbit

* **New Features**
  * Per-pattern DLP action "warn": treats matches as informational, allows traffic, and reports them separately from enforced matches.
  * Warn-mode emits distinct audit/log events for monitoring and provides hook points for integrations.

* **Documentation**
  * New guide for tuning DLP false positives using warn mode.
  * Configuration docs updated; built-in default patterns cannot be set to warn.

* **Tests**
  * Expanded coverage for warn vs enforce behavior, reload semantics, and hook/emission paths.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -324,7 +324,7 @@ dlp:
 | `patterns` | 46 built-in | DLP credential detection patterns |
 | `patterns[].validator` | `""` | Post-match checksum validator: `luhn`, `mod97`, `aba`, or `wif` |
 | `patterns[].exempt_domains` | `[]` | Domains where this pattern is not enforced (wildcard supported) |
-| `patterns[].action` | `""` | Per-pattern action override. Only `warn` is supported. When set to `warn`, matches produce an audit event (`dlp_warn`) without blocking. See the [false positive tuning guide](guides/false-positive-tuning.md) for the rollout workflow. Built-in default patterns cannot be set to warn. |
+| `patterns[].action` | `""` | Per-pattern action override. Only `warn` is supported. When set to `warn`, matches allow traffic through without enforcement. See the [false positive tuning guide](guides/false-positive-tuning.md) for the rollout workflow. Built-in default patterns cannot be set to warn. |
 
 There is no top-level `dlp.action` setting. DLP enforcement is transport-specific:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -324,6 +324,7 @@ dlp:
 | `patterns` | 46 built-in | DLP credential detection patterns |
 | `patterns[].validator` | `""` | Post-match checksum validator: `luhn`, `mod97`, `aba`, or `wif` |
 | `patterns[].exempt_domains` | `[]` | Domains where this pattern is not enforced (wildcard supported) |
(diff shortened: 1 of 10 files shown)


luckyPipewrench:
### GPT-5.4 review triage

**1. Package-level DLPWarnHook (medium):** Agree the instance-level hook or interface injection is the right long-term pattern. The global exists because `run.go` — where Scanner construction and audit logger wiring happen — is reserved by #384. The follow-up (already tracked in state.md) will move the hook onto the Scanner when `run.go` is available. The atomic-swap concern doesn't apply today since the hook is set once at startup and never changed.

**2. deduplicateWarnMatches key (medium):** URL DLP warn matches intentionally dedup by pattern name only. The same pattern matching multiple URL components (path, query, hostname collapse) is one logical finding — operators want "pattern X matched this request" not "pattern X matched 4 different query decompositions." Text DLP matches carry richer provenance (encoding field) and dedup at the TextDLPMatch level via the existing `deduplicateMatches` which keys on `(name, encoding)`.

**3. context.Background() in LogDLPWarn (low):** This follows the existing convention — all 20+ `Emit` calls in `logger.go` use `context.Background()`. Emit is fire-and-forget by design (async buffered channel, UDP syslog). Changing this for one method would be inconsistent. The CodeRabbit review raised the same point about the hook signature lacking request context; both are tracked for the follow-up.

DeathCamel58 (dismissed):

DeathCamel58 (approved):

PR #403, opened 2026-04-16
feat: RFC 9421 envelope signing + canonical policy hash + redirect refresh

luckyPipewrench:
## Summary

Adds RFC 9421 HTTP Message Signatures to the mediation envelope, makes the policy hash (`ph`) admission-grade, and rebuilds the envelope on redirect hops.

- **RFC 9421 envelope signing.** Opt in via `mediation_envelope.sign: true` and an Ed25519 `signing_key_path`. The pipelock signature uses the `pipelock1` dictionary label and the `pipelock-mediation` tag, so it coexists with any upstream `sig1` / Web Bot Auth signature on the same request. Per-request component list is dynamic — `content-digest` is dropped on body-less GETs and on requests whose body exceeds `max_body_bytes`. New config fields: `sign`, `signing_key_path`, `key_id`, `signed_components`, `created_skew_seconds`, `max_body_bytes`. Supported components are restricted to `@method`, `@target-uri`, `@authority`, `content-digest`, `pipelock-mediation`; typos and unsupported entries are rejected at config validation and signer construction.
- **Canonical policy hash.** `ph` (first 16 bytes of the dictionary key) now derives from a canonicalised, slice-order-preserving JSON projection of the effective config. Whitespace, comments, and reordering noise fields no longer shift `ph`; behavioural rule reorders (DLP patterns, MCP tool policy rules, chain rules, suppress entries) still do. Per-agent resolved configs each compute and cache their own canonical hash.
- **Redirect refresh.** `CheckRedirect` rebuilds the Pipelock-Mediation header on every allowed redirect through the fetch and forward proxy: increments `hop` (new wire key, omitted when zero), recomputes method-bound action/side-effect, drops stale `Content-Digest`, re-signs over the new `@target-uri`. The hop counter uses `len(via)` so it tracks across multi-hop chains, not just the first refresh.

## Fail-closed contract

Every transport that emits a signed envelope fails closed on a sign error: forward proxy, intercepted CONNECT, fetch handler, reverse proxy (via a `http.RoundTripper` wrapper that runs *after* `httputil.ReverseProxy`'s Director rewrites the URL, so `@target-uri` matches what the upstream actually receives), WebSocket handshake (signed at the dial point with a synthesised `*http.Request`), and the redirect refresh. Sign errors return 403 / WS policy-violation close, log a block receipt, and never forward unsigned or partially-mutated requests.

Reload semantics are symmetric for both signing keys (envelope + receipt): if either key is unreadable on reload, the entire config swap aborts, the previous emitter and its in-memory key stay installed, and the proxy continues serving on the last good config until ops fixes the file. There is no silent downgrade to unsigned, and a broken envelope reload cannot partially advance receipt state for a config that never activated.

## Inbound strip bug fix

`envelope.StripInbound` previously used `strings.Split(val, ",")` to tokenise `Signature` / `Signature-Input` dictionary members. Under RFC 8941, dictionary members may carry commas inside quoted parameter values; the naive split corrupted surviving members and left dictionary residue that no longer parsed as a structured field — a sanitisation bypass for attacker-crafted inbound signature headers. The strip now parses all header values as a single RFC 8941 dictionary via `httpsfv`, drops `pipelock*` members, and re-serialises. Unparseable dicts fail closed (entire header dropped).

## Transport coverage

| Transport | Envelope | Signing | Where signing runs | Redirect refresh |
|-----------|----------|---------|--------------------|------------------|
| Fetch proxy | yes | yes | pre-dispatch | yes |
| Forward proxy | yes | yes | pre-dispatch | yes |
| Intercepted CONNECT | yes | yes | pre-dispatch | n/a (one-shot RoundTrip) |
| Reverse proxy | yes | yes | RoundTripper wrapper (post-Director) | n/a (redirects disabled) |
| WebSocket handshake | yes | yes | pre-dial via synthesised request | n/a |
| MCP stdio | yes | no (HTTP-only) | pre-dispatch | n/a |
| MCP HTTP / SSE | yes | no | pre-dispatch | n/a (redirects disabled) |

MCP stdio signing is explicitly out of scope — RFC 9421 is defined on HTTP messages and the stdio transport rides the envelope in JSON-RPC `_meta`. MCP HTTP transports return `http.ErrUseLastResponse` from `CheckRedirect` (SSRF hardening) so the redirect refresh helper is documented as N/A there; if a future change enables redirect following on those paths, the helper must be wired in.

## Inbound verification

Out of scope for this PR. `internal/envelope/inject.go` carries a `TODO(v2.2.x): inbound verification + replay cache` marker. Pipelock continues to strip any inbound `pipelock*` signature on the assumption it is forged or stale.

## Documentation

`docs/mediation-envelope.md` is new in this PR. It covers the envelope wire format, the canonical policy hash derivation, the signing config surface, the supported component allowlist, dynamic component-list semantics, coexistence with upstream signatures, reload semantics, the transport coverage matrix, redirect refresh, capability separation requirements for the signing key, and the v1 known limitations (MCP stdio out of scope, inbound verify deferred, Ed25519-only, RFC 9530 SHA-256 vs library SHA-512 digest mismatch on the body-bearing interop path, blank `aa` on anonymous requests, `max_body_bytes: 0` is treated as "use the default").

## Test-only dependency

`github.com/common-fate/httpsig` is added to `go.mod` for use inside `internal/envelope/interop_test.go` only. It is imported from a `_test.go` file and never reaches production binaries. Two cross-library interop tests run against pipelock's signer:

- `TestRFC9421_ExternalVerifierInterop` parses a pipelock-signed request through `common-fate/httpsig`'s `verifier.Verifier` with an `alg_ed25519.SingleKeyDirectory`.
- `TestRFC9421_ReferenceVerifierInterop` reconstructs the RFC 9421 §2.5 signature base from scratch using only `httpsfv` (already vendored) and runs `crypto/ed25519.Verify` directly.

The body-bearing interop path is exercised by pipelock's own per-transport tests; the cross-library test uses a body-less GET because the verifier library hard-codes SHA-512 for content-digest while pipelock follows the RFC 9530 SHA-256 default. Resolving the digest-family mismatch is tracked as a v2.2.x follow-up.

## Test plan

CI covers everything — race detector, `-tags enterprise`, lint, build, govulncheck, CodeQL, self-scan. No manual reviewer steps required.

## Summary by CodeRabbit

* **New Features**
  * Optional Ed25519 HTTP Message Signatures for mediation envelopes with configurable signing options
  * Mediation header refresh and re-signing on allowed proxy redirects; stale content-digest dropped

* **Security Hardening**
  * Canonical policy-hash derived from a deterministic canonical config projection
  * Strict structured-field parsing for signature headers and fail-closed signing/refresh behavior

* **Documentation**
  * New mediation envelope guide covering format, signing, redirect behavior, and known limitations

* **Tests**
  * Extensive new unit and integration tests covering signing, envelope lifecycle, redirects, and fail-closed paths


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #408, opened 2026-04-17
fix: v2.2.0 pre-tag hardening bundle

luckyPipewrench:
## Summary

Consolidated fix arc from the v2.2.0 pre-tag adversarial gate (13-row canonical gate harness, 11 rounds). Every item is a small self-contained fix for a real observed defect; grouping keeps the release-branch narrative coherent.

## Scope

**Forward proxy / TLS-intercept header hygiene**
- Strip `X-Pipelock-Agent` header + `?agent=` query param on the forward and TLS-intercept paths so caller-supplied identity never bleeds through when `bind_default_agent_identity` is set.
- Strip the full forwarded-IP family (`X-Forwarded-For`, `X-Real-IP`, `X-Forwarded-Host`, `X-Forwarded-Proto`, `X-Forwarded-Port`, `Forwarded`, `Via`) on both paths.
- Three new integration tests in `internal/proxy/intercept_test.go` pin strip + canary body-block behaviour end-to-end on the TLS-intercept path.

**MCP subprocess teardown (defense-in-depth)**
- Direct child in its own process group (`Setpgid=true`); SIGTERM/SIGKILL the pgid after `Wait`.
- Context-cancellation watcher goroutine proactively SIGTERMs the negated pgid on parent cancel, beating the post-`Wait` SIGKILL backstop.
- Linux `setPdeathsig` helper sets `SysProcAttr.Pdeathsig = syscall.SIGTERM` so the kernel TERMs the direct child if pipelock dies.
- `/proc`-walk subreaper sweep after the pgid kill catches `setsid`/double-fork descendants reparented via `PR_SET_CHILD_SUBREAPER`.
- Pipelock-`SIGKILL` (uncatchable) still orphans cooperative grandchildren to PID 1 — structural and documented as v2.2.1 deployment-guidance work.

**Network-exfil tool policy**
- Broaden curl upload-flag pattern to cover `-F`, `--form`, `--data-binary`, `--data-raw`, `--data-urlencode`, `--post-file`, `--body-data`, `--body-file` across all six preset configs.

**Recorder robustness**
- Startup writability probe refuses to start when the evidence dir is unwritable.
- `ensureFile` stats the dir on every call, recovers from mid-run `rm -rf evidence/` by recreating the dir and dropping the stale fd; loud stderr notice per recovery.
- Hot-reload rejects `flight_recorder.signing_key_path` changes with a clear warning and preserves the previously-loaded key. Key rotation requires a restart (chain-rotation markers deferred to v2.2.1).

**Config reload coalescing**
- Short-circuits duplicate-hash reloads so fsnotify+SIGHUP stacking fires once per mutation, while preserving single-SIGHUP acks for operators.

**Sandbox best-effort posture**
- Loud `WARNING` line alongside `DEGRADED` on both `pipelock sandbox --best-effort` and the standalone init path used by `pipelock mcp proxy --sandbox-best-effort`, so the proxy-bypass risk surfaces at startup.

**Audit-log and client-response URL redaction**
- URLs for content-matching scanner blocks (`dlp`, `body_dlp`, `header_dlp`, `mcp_input_scanning`, `response_scan`, `address_protection`, `seed_phrase`, `cross_request_entropy`) truncate to scheme+host in structured logs AND in the client-facing fetch 403 body. Previously the raw secret that triggered the match was echoed in both.

**MCP block-response wording**
- `blockResponseReason` helper gives accurate wording on the tool-poisoning block path (tools/list) and provenance-verification path. Previously reported as "prompt injection detected" — a false scanner-attribution that wasted debug time.

**Envelope inbound strip precision**
- Inbound `Signature`/`Signature-Input` strip fires only when raw bytes contain the pipelock member prefix, so legitimate Web Bot Auth `sig1` members whose values are not strict-base64 survive. Partial-parse mixing (pipelock-poisoned bytes alongside a valid `sig1`) is a known fail-closed drop pending an RFC 8941 partial parser.

**Install subcommand**
- New hidden `pipelock install <dest>` reads its own executable and writes it atomically to the destination, so scratch-based sidecar init containers can populate a shared-bin volume without a shell. Sidecar-mate CLI now matches the sidecar proxy byte-for-byte.

## Test plan

Exercised by the canonical 13-row pre-tag gate harness over 11 rounds ending in `9 PASS + 4 SKIP + 0 REGRESS`. Test suite runs `-race -count=1` clean; `golangci-lint run ./...` clean on the full repo.


luckyPipewrench:
All three addressed in `1c663cf3`.

**#1** `containsPipelockMember` is now case-insensitive `strings.Contains(lower, "pipelock")` — OWS/tab/line-fold variants no longer smuggle a malformed pipelock member past the fail-closed drop. Added `TestStripInbound_MalformedPipelockFormattingVariants` covering the variants. Well-formed `pipelockless=...` still survives via the httpsfv branch.

**#2** `installCmd` now `Lstat`s the destination (rejects symlinks and non-regular files) and writes via `atomicfile.Write` for temp-file-rename semantics. Went with atomic-rename over `O_NOFOLLOW` because it also covers partial-write safety. New `install_test.go` pins both reject paths.

**#3** Captured `childPgid` at `cmd.Start` (verified against `syscall.Getpgid`); both the `ctx.Done` watcher and post-`Wait` cleanup signal the captured value. Skipped the per-kill re-verification and the routine teardown audit log — capture-at-Start already closes the recycle race, and audit-logging every MCP subprocess exit would drown operators in close-time noise.

luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #413, opened 2026-04-19
feat: class-preserving redaction library and config schema

luckyPipewrench:
## Summary

New `internal/redact` package provides class-preserving, irreversible redaction of secrets in JSON request bodies bound for upstream providers. Typed placeholders `<pl:CLASS:N>` preserve semantic class without exposing originals.

19 built-in classes (IPs, FQDN, email, AWS/Google/GitHub/Slack tokens, JWT, hashes, SSH key header, AD user, SSN, credit card) plus operator YAML dictionaries. Whole-body JSON walk — every string scalar and every object key at any depth, no field-position exemption. Fail-closed on oversize, unparseable, overflow, depth-exceeded with typed `*BlockError` reasons.

Adds `ActionRedact = "redact"` and a `Redaction` config block plumbed into `Config`. Library ships disabled.

## Scope

Library-only. No transport calls it yet — forward-proxy hook and receipt integration land in a follow-up.

## Rationale

Whole-body scan (no `system`/`tools` exemption) is deliberate: the caller controls field placement, so any field-position exemption is a bypass.

## Release Notes

* **New Features**
  * Added redaction framework to secure sensitive data in request bodies. The feature is currently disabled and enforced as disabled in this release. Configuration now supports profiles, dictionaries, and limits to detect and replace secrets (API keys, tokens, credentials, hashes, etc.) with safe placeholders.

* **Tests**
  * Added comprehensive test coverage for redaction functionality including configuration validation, secret detection, and JSON processing.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -26,6 +26,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/luckyPipewrench/pipelock/internal/envelope"
+	"github.com/luckyPipewrench/pipelock/internal/redact"
 	"github.com/luckyPipewrench/pipelock/internal/signing"
 )
 
@@ -59,6 +60,10 @@ const (
 	ActionStrip    = "strip"
 	ActionForward  = "forward"
 	ActionAllow    = "allow"
+	// ActionRedact replaces the matched value with a typed placeholder
+	// (e.g. "<pl:ipv4:1>"). Irreversible: pipelock holds no mapping.
+	// See the redaction-v1 design spec in ops for the semantic model.
+	ActionRedact = "redact"
 )
 
 // Severity constants for chain detection and emit thresholds.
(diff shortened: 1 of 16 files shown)


luckyPipewrench:
Addressed all four findings from `/review deep`:

**#1 Enforcement gap.** `Validate()` in `internal/config/config.go` now blocks startup when `redaction.enabled: true` until the request pipeline enforces the feature. The library-level `redact.Config.Validate()` stays permissive so internal callers and tests can still exercise the schema — the pipelock gate is what operators hit.

**#2 Key collision.** Walker detects rewritten-key collisions in `map[string]interface{}` and fails closed with `ReasonKeyCollision` instead of the later-write-wins silent drop.

**#3 Placeholder class safety.** `Redactor.Placeholder` coerces any class that doesn't match the safe shape (`[a-z0-9][a-z0-9_-]*`) to the sentinel class `invalid`. Downstream parsers never see a perturbed `<pl:CLASS:N>` even if a caller bypasses `AddDictionary`'s class validation.

**#4 Allowlist validation.** `allowlist_unparseable` entries are now canonicalised at config load: reject scheme, path, port, uppercase, leading/trailing dot, consecutive dots, and invalid characters. Match semantics for v1b will be unambiguous.

Regression tests added for each.

DeathCamel58 (changes requested):
Please audit the codebase for other redaction code, and migrate to use the redact package. No point in maintaining multiple packages that do the same thing essentially.

DeathCamel58 (approved):

PR #422, opened 2026-04-22
refactor: consolidate runtime config resolution into Config.ResolveRuntime

luckyPipewrench:
Adds `config.ResolveRuntime` as the single entrypoint that produces a runtime `*Config` from a loaded one: deep clone, MCP proxy response-scanning fallback, caller-supplied bundle merge, and mode-aware MCP scanning auto-enable. `pipelock run`, `pipelock mcp proxy`, `pipelock mcp scan`, and the hot-reload goroutine all call it once and feed the resolved clone to every downstream consumer (scanner, emitters, proxy, reload diff). Also adds repo-local hardening audits so the new invariants cannot regress.

## Motivation

- MCP input scanning, tool scanning, and tool policy auto-enable lived in three places with the same semantic rule.
- Bundle merging, response-scanning fallback, and auto-enable mutated the loaded config in separate blocks scattered through the runtime. One builder keeps the order explicit.
- Per-agent effective configs and reload diffing both benefit from a resolved-config abstraction.
- Hot reload previously compared pre-resolve `newCfg` against post-resolve `oldCfg`, which produced false downgrade warnings on no-op reloads. Reload now resolves first, then compares resolved-vs-resolved.

## Hash split

Receipt `ConfigHash` continues to use `cfg.Hash()` (raw YAML audit fingerprint). Envelope `ConfigHash` uses `cfg.CanonicalPolicyHash()` (policy-semantic attestation fingerprint). Call-site comments document the intentional split: receipts are a point-in-time audit fingerprint of the loaded file, while envelope `ph` is the effective-policy fingerprint that verifiers care about.

The canonical hash cache is now stored as `*canonicalHashCacheHolder` (pointer to `atomic.Value`) rather than an embedded value, so `Config.Clone`'s shallow struct copy duplicates the pointer rather than the atomic. `CanonicalPolicyHash` lazily allocates a fresh holder if none exists and `Clone` explicitly reassigns a fresh one — both paths avoid Go's atomic-copy-after-first-use rule.

## Modes

| Mode | Auto-enable MCP sections | Response-scanning fallback |
|------|--------------------------|----------------------------|
| `RuntimeForward` | no | no |
| `RuntimeForwardWithMCPListener` | yes | no |
| `RuntimeMCPProxy` | yes | yes |
| `RuntimeMCPScan` | no | yes |

`WrapsMCP()` and `NeedsResponseScanningFallback()` capture the predicates so both the resolve builder and any future caller stay in sync.

## Release hardening

Adds repo-local checks so release and high-risk runtime paths have explicit, repeatable gates:

- `.github/workflows/hardening.yaml` runs `make release-audit` and `make test-runtime-critical` as blocking checks and `make runtime-policy-audit` plus a complexity/duplication debt lint as warning-only reports summarised in the step summary.
- `scripts/release-audit.sh` blocks `pull_request_target`, `permissions: write-all`, unpinned actions, `persist-credentials: true`, PR-triggered workflows that reference non-`GITHUB_TOKEN` secrets, and `issue_comment` workflows that either use secrets without an `author_association` gate or check out the PR head ref instead of the merge ref.
- `scripts/runtime-policy-audit.sh` scans `internal/cli/runtime`, `internal/mcp`, and `internal/proxy` for direct assignment of policy-relevant config fields (`MCPInputScanning`, `MCPToolScanning`, `MCPToolPolicy`, `ResponseScanning`, `DLP.Patterns`, `Internal`) outside of tests. Locks in the resolve builder so the mutation pattern cannot regress.
- Makefile targets: `test-runtime-critical` (race detector on `config`, `cli`, `mcp`, `proxy`), `release-audit`, `runtime-policy-audit`, `debt-check`, `release-check` (aggregate gate).
- The tag release workflow now runs the audits before tagging.
- `docs/release-checklist.md` captures the release gate explicitly.

## Tests

`internal/config/runtime_test.go` covers immutability, deep-clone aliasing, hash semantics, resolve-info reporting, response-scanning fallback ordering, mode helpers, and the resolved-vs-resolved reload comparison. `internal/cli/cli_test.go` adds two end-to-end reload tests (MCP-listener reload warnings, strict-mode reload with a non-policy change). `internal/cli/runtime/resolve_logs_test.go` covers the operator-facing log helper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hot-reload validation to prevent security downgrades during strict-mode config changes.
  * Fixed kill-switch token behavior so API status follows updated config after reloads.

* **Improvements**
  * Emit clearer notices when runtime defaults are auto-enabled (distinguished by runtime mode).
  * Runtime config handling made immutable for safer, deterministic reloads and receipt fingerprints.

* **Tests & Hardening**
  * Added comprehensive unit/integration tests for runtime resolution and reloads.
  * New CI hardening workflow, release audit scripts, Make targets, and a release checklist document.

### DIFF
--- a/.github/workflows/hardening.yaml
+++ b/.github/workflows/hardening.yaml
@@ -1,0 +1,101 @@
+name: Hardening
+
+on:
+  push:
+    branches: [main]
+    tags-ignore: ['v*']
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: hardening-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+permissions:
+  contents: read
+
+jobs:
+  workflow-audit:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Workflow and release audit
+        run: make release-audit
+
+  runtime-critical:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version: '1.25'
+
+      - name: Verify dependencies
+        run: go mod verify
+
+      - name: Download dependencies
+        run: go mod download
+
+      - name: Run runtime-critical packages
+        run: make test-runtime-critical
+
+  hardening-report:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        with:
+          go-version: '1.25'
+
+      - name: Runtime policy audit (warning-only on PRs; blocking in release workflow)
+        id: policy
+        continue-on-error: true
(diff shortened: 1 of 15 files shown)


luckyPipewrench:
@coderabbit review

DeathCamel58 (approved):

PR #442, opened 2026-04-26
feat(contract): v2.4 schema package + EvidenceReceipt v2 envelope

luckyPipewrench:
## Summary

- Land `internal/contract` schema package: Contract, ActiveManifest, CompileManifest, Tombstone, KeyRoster, VerificationMetadata.
- Land `internal/contract/receipt` package: EvidenceReceipt v2 envelope with a 13-payload registry, strict per-payload decoding, key_purpose authority matrix at the envelope level, and v1-dispatch rejection.
- JCS canonicalization (RFC 8785) over typed Go structs with NFC normalization, duplicate-key rejection, NFC-collision rejection, float rejection, and trailing-token rejection.
- Strict YAML loader rejects duplicate keys, anchors, aliases, merge keys (`<<:`), custom tags (including the long-form `!<tag:yaml.org,2002:int>`), and multi-document streams.
- Ed25519 PureEdDSA verify (RFC 8032 section 5.1.6 only; no Ed25519ph or ctx variants) plus key_purpose authority matrix.
- Binary SHA-256 Merkle root over canonical rule bodies with leaf/internal domain separation and odd-count duplication.
- Schema-only. No inference, runtime, or storage. Foundation for the v2.4 learn-and-lock work.

## Validation surface

Each typed artifact has a `Validate()` that enforces:
- `schema_version` against the current value.
- enum fields (manifest_kind, contract_kind, bundle_kind, validation_outcome, request_kind).
- selector_id is recomputed from the JCS preimage and matched against the claimed value (no identity forgery).
- selector_set_hash is required and recomputed from sorted selector_ids.
- module_digest_root is recomputed from the JCS-canonicalized module set.
- tombstone_index_root is recomputed from the sorted tombstone hash list.
- compile-manifest `settings` is restricted to the documented allowlist (`confidence`, `normalization`, `shadow`, `drift`, `privacy.*`, `redaction.*`); scalar values for structured sub-keys reject.
- `data_class_root` is enum-validated and `regulated` is rejected.
- recursive data-class coverage; `regulated` fields in any signed body reject.
- EvidenceReceipt v2 envelope enforces signature shape (algorithm, `ed25519:` prefix, 64-byte hex), required signer_key_id and key_purpose, and per-payload-kind authority (for example, `proxy_decision` must be signed by `receipt-signing`).

## Cross-implementation conformance

Golden vector fixtures under `internal/contract/testdata/golden/` cover every signed-artifact type, signed with the published RFC 8032 section 7.1 test key. An independent Python verifier at `testdata/python_verifier_fixture/verify.py` recomputes the JCS preimage and verifies each signature using `cryptography` and `jcs`. The round-trip proves byte-identical canonicalization between Go and Python.

## New direct dependency

`github.com/goccy/go-yaml` is added for the strict YAML loader. It is required because it exposes both AST-level anchor/alias/tag detection and duplicate-key rejection, which `gopkg.in/yaml.v3` does not. Existing scanner and proxy YAML paths are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Major expansion of the contract subsystem: manifests, compile/activation flows, deterministic canonicalization, Merkle roots, data-class rules, tombstones, key rosters, verification metadata, and typed signable preimages.
  * Enhanced evidence-receipt system with many typed payloads, strict validation and signature verification; Python verifier CLI for cross-language golden checks.

* **Tests**
  * Large test suite additions: golden-vector fixtures, cross-language verifier integration, and extensive strict JSON/YAML unit tests.

* **Documentation**
  * Package docs and Python verifier README added.

* **Chores**
  * CI/pre-commit config tweaks and YAML parsing dependency added.


luckyPipewrench:
Pushed `2fa1c9cd` addressing the latest review pass.

**Security fixes (both MAJOR):**

- `internal/contract/load.go:36`: trim ASCII JSON whitespace before the `string(raw) == "null"` compare. Closes the bypass where ` null`, `\tnull\n`, etc. slipped past the guard and let `json.Decoder.Decode` bind null to a typed struct as the zero value with no error. Test grew from 2 cases to 10 covering every leading/trailing/wrapped whitespace permutation.
- `testdata/python_verifier_fixture/verify.py:64`: reject `"signatures": []` with `empty signatures array` before the verify loop. Previously an empty list silently returned `(True, "")` because the loop never executed. The Go side never emits an empty list today, but this script is the cross-implementation oracle and must fail closed.

**Doc clarification (deep-review medium):**

- `internal/contract/contract.go:113`: doc comment on `Contract.Validate` now names itself the second-line check on the typed struct. Unknown-field rejection lives at the transport boundary in `LoadContract` / `DecodeStrictJSON`. Callers that bypass `LoadContract` forfeit unknown-field detection. No code change because the strict-decode invariant is already enforced at the loader; the fix is making the layering explicit so callers cannot skip it by mistake.

luckyPipewrench:
Pushed `47d8a1c9`. cryptography pin bump.

Verified the three advisories CodeRabbit flagged.

- GHSA-r6ph-v2qm-q3c2 (HIGH) SECT-curve subgroup attack. Fixed in 46.0.5.
- GHSA-p423-j2cm-9vmq (MODERATE) non-contiguous buffer overflow. Fixed in 46.0.7.
- GHSA-m959-cc7f-wv43 (LOW) incomplete DNS name constraint enforcement. Fixed in 46.0.6.

Floor raised from `>=42.0.0,<46.0.0` to `>=46.0.7,<47.0.0` to land all three patches.

Verified end to end. A fresh `pip install 'cryptography>=46.0.7,<47.0.0' 'jcs>=0.2.1,<1.0.0'` plus `go test -run TestGoldenVectors_PythonVerifier` produces 7/7 OK against the golden vectors. Ed25519 PureEdDSA verify is API stable across 42.x to 46.x so the bump is drop in.

luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #443, opened 2026-04-26
fix(proxy/reload): scanner drain, detect_drift rebuild, reverse-proxy receipt parity

luckyPipewrench:
## Summary

Three reload-and-receipt fixes that share a review surface.

### Scanner drain on close

`scanner.Scanner` gains `BeginUse` / `Closed` and a drain-on-`Close` pattern (closeMu RLock plus WaitGroup, capped at 30 seconds). `Proxy.Reload` runs the prior scanner's `Close` in a goroutine so reload returns promptly while drain blocks teardown until in-flight scans release. Today only ticker goroutines stop on `Close`, but future destructive resources (sqlite handles, file descriptors) get the safety guarantee for free. `ReverseProxyHandler.snapshotAndAcquire` registers in-flight use atomically; on a reload race it re-loads the swapped-in scanner up to three times before falling through with a no-op release.

### detect_drift rising-edge baseline rebuild

`mcp_session_binding.detect_drift` flipping `false` to `true` via hot reload left the per-listener `toolBaseline` populated with hashes from before the disabled window, so post-flip drift detection ran against stale ground truth. Adds `tools.ToolBaseline.ResetDriftState` (clears `hashes` / `descs` / `params`, preserves `knownTools` / `hasBaseline`) plus a small `tools.DetectDriftRisingEdge` helper consumed by both the per-listener (`internal/mcp/proxy_http.go`) and server-level (`internal/cli/runtime/server.go`) closures so the rising edge re-seeds drift state on the next `tools/list`. Other transitions (true to true, true to false, false to false) are no-ops.

### Reverse-proxy block-path receipt parity

Reverse-proxy SSE-stream findings, compressed-body fail-closed blocks, oversize-body blocks, and read-error blocks now emit signed action receipts under `Transport: "reverse"`, reaching parity with forward and intercept on the same classes of block. New `ReverseProxyHandler.SetReceiptEmitter` mirrors `SetEnvelopeEmitter`; one `ActionID` per response covers all four guards. New layer constant `LayerReverseResponseBlocked` covers the non-finding fail-closed paths (compressed, oversize, read-error); SSE keeps the existing `LayerSSEStream` shared across transports.

## Receipt parity matrix

| Transport | SSE block | Compressed block | Oversize block | Read-error block |
|-----------|-----------|------------------|----------------|------------------|
| forward   | receipt   | log+metric only* | n/a            | n/a              |
| intercept | receipt   | receipt          | receipt        | receipt          |
| fetch     | n/a       | receipt          | receipt        | n/a              |
| reverse   | receipt (new) | receipt (new) | receipt (new) | receipt (new) |

\* The forward-proxy non-SSE compressed block is a separate gap surfaced during this audit. It stays out of scope here so the PR remains focused on the reverse-proxy parity.

## Hot-reload state matrix

`Proxy.Reload` is exercised across the five reload states from CLAUDE.md (first load, first reload, second unrelated reload, downgrade, idempotent reload). After every reload the prior scanner reaches `Closed=true` without affecting the live scanner that handles new traffic.

## Summary by CodeRabbit

* **Documentation**
  * Updated static binary size estimates from ~18MB to ~20MB across guides and comparisons.

* **Bug Fixes / Reliability**
  * Hot-reload drift handling now resets baseline when detection is re-enabled to avoid stale tool inventories.
  * Scanner shutdown now waits for in-flight scans to drain and fails requests safely during reload races (returns 503 when scanner unavailable).
  * Reverse-proxy now consistently emits receipts for blocked/fail-closed cases (compressed, oversized, read-error, SSE).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,7 +30,7 @@ These must be proven by tests, not assumed from docs or deployment.
 | Module | `github.com/luckyPipewrench/pipelock` |
 | Go | 1.25+ (CI tests 1.25 and 1.26) |
 | License | Apache 2.0 (core), ELv2 (`enterprise/`) |
-| Binary | Single static binary, ~18MB |
+| Binary | Single static binary, ~20MB |
 | Deps | 21 direct deps — run `make stats` for the live count. Core set: cobra, zerolog, go-readability, yaml.v3, prometheus, fsnotify, gobwas/ws, sentry-go, modernc.org/sqlite, otlp/proto, google/protobuf, go-landlock, cyclonedx-go, google/uuid, common-fate/httpsig (RFC 9421), dunglas/httpsfv (RFC 8941), plus x/crypto, x/net, x/sys, x/text, x/time. |
 
 ## Docs & README Messaging
--- a/docs/comparison.md
+++ b/docs/comparison.md
@@ -8,7 +8,7 @@ An honest feature matrix and guidance on when to use what.
 |---------|----------|-----|---------|-----|
 | **Layer** | Application firewall + process containment (HTTP + MCP + WebSocket + Landlock + seccomp + netns) | MCP proxy | Kernel (seccomp/eBPF/FUSE) | OS sandbox |
 | **Language** | Go | Go | Go | TypeScript |
-| **Binary** | Single, ~18MB | Single | Single + kernel modules | npm package |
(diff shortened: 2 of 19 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #452, opened 2026-04-28
feat(contract): add inference confidence and exposure gates

luckyPipewrench:
## Summary

Adds the `internal/contract/inference` package: pure Go statistical primitives for classifying behavioral-rule confidence from recorded observation events. Wires the deployment-configurable surface (exposure floors) into config, and the operational surface (verdict and floor-failure histograms) into the metrics bundle. No runtime wiring yet; that lands in a follow-up alongside `pipelock learn compile`.

## What's in the package

- **`wilson.go`**: Wilson score interval lower and upper bound for binomial proportion confidence on small or skewed samples (Newcombe 1998 closed form). The 0.95 z-score is fixed for the production path so identical recorder data uses the same statistical contract across deployments; the alpha-parameterized signature stays for testability. `DefaultWilsonAlpha = 0.05` is a named package constant rather than a config field. The confidence level is part of the statistical contract, and making it deployment-configurable would let two installs infer different classifications from identical recorder data.

- **`opportunity.go`**: Conditional-opportunity denominator picker per the contract design's hierarchy table (host stability to sessions, path-family stability to host sessions, method stability to host/path-family events, header shape / arg schema to tool-endpoint calls, sequence n-gram to source-action windows). Strict per-load validation at the boundary; unknown level rejects with `ErrUnknownOpportunityLevel`, negative input rejects with `ErrNegativeOpportunityCount`.

- **`floors.go`**: Three exposure-floor gates (min_sessions, min_events, min_windows, defaults 5 / 20 / 3) plus the `Classify` composer that AND-composes Wilson plus floors into a `never_confirmed` / `brittle` / `stable` verdict. Floors are deployment-configurable (traffic volumes differ across deployments); thresholds (tau_brittle = 0.50, tau_stable = 0.85) are hardcoded. Any floor failure forces `never_confirmed` regardless of Wilson, proven by the 8-row floor-pass/fail by Wilson-stable matrix test.

- **`budgets.go`**: p99, p95, median, max, sample_count statistics for numeric rate and size budgets, with caller-selectable headroom (`DefaultHeadroomRate = 0.20`, `DefaultHeadroomSize = 0.50`). Nearest-rank percentile via integer arithmetic, keeping results deterministic without floating-point ceil-rounding edges. Empty input returns a zero-valued Budget (no panic). `ThinSample(minEvents)` flag for the review-UX badge.

## Config surface

Adds `learn.inference.floors.{min_sessions, min_events, min_windows}` to the existing `learn:` block.

- Missing or zero resolves to defaults applied at runtime via `inference.Floors.Resolved()`.
- Negative is rejected at config-load with a precise YAML field-path error.
- The config validator deliberately duplicates the inference package's `Floors.Validate()` so operators see `learn.inference.floors.min_sessions: -1: must be non-negative` instead of an API-level field name. Layering rationale documented inline.

## Metrics surface

Two new counters in the existing `internal/metrics/learn.go` bundle (no new file, per-feature bundling stays intact):

- `pipelock_learn_inference_classify_total{outcome}`: outcome is one of `never_confirmed`, `brittle`, `stable`. Powers the review-UX verdict histogram.
- `pipelock_learn_inference_floor_failures_total{floor}`: floor is one of `sessions`, `events`, `windows`. Diagnostic for which floor is the bottleneck on a deployment's data volume.

Nil-safe recorder helpers parallel to the existing observation-pipeline pattern.

## Policy-hash impact

Adding floors to the `Learn` config flows into the canonical policy hash. Floors are detection-relevant: they decide stable vs never_confirmed at compile time, so verifiers must observe the schema change to detect deployments that loosen the exposure gates. Both `goldenHashDefaults` and `goldenHashRichConfig` shift in lockstep with `Defaults()`. Deliberately hash-pinned so the verifier's authenticated state matches the contract.

## Coverage and tests

- 100% line coverage on every function in `internal/contract/inference/` (wilson, opportunity, floors, budgets); 18 of 18 functions at 100%.
- 100% on the new config validator (`validateLearnInferenceFloors`) and on every new metrics helper (`RecordInferenceClassification`, `RecordInferenceFloorFailure`).
- 16-corner-case Wilson test pack matching the design table to 3 decimal places, plus property tests for monotonicity, determinism over 1000 invocations, and complementary symmetry of upper/lower bounds.
- 8-row floor-pass/fail by Wilson-stable matrix proving floor failure overrides Wilson regardless of confidence.
- `p99 >= p95 >= median`, `max >= p99` properties verified across 5 different windows.

## Cross-build

Linux and Windows, default and enterprise build tags, all four green. Inference is pure Go math plus `slices` stdlib, no syscall surface.

## Out of scope

- Runtime wiring (the inference engine consuming live recorder events): follows in a later PR.
- `pipelock learn compile` end-to-end CLI: follows in a later PR.
- Path-family normalization and cardinality cap: follows in a later PR.
- The verifier `verify.py` is not extended; this PR does not emit signed artifacts.

## Release Notes

* **New Features**
  * Added inference-based rule classification system with confidence levels (stable, brittle, never confirmed) to indicate rule reliability
  * Configurable exposure thresholds for sessions, events, and windows to control when rules are classified as stable
  
* **Chores**
  * Added new metrics to monitor rule classifications and exposure threshold failures

### DIFF
--- a/internal/config/canonical_golden_test.go
+++ b/internal/config/canonical_golden_test.go
@@ -37,7 +37,12 @@ const (
 	// Defaults() now includes the Learn top-level block (enabled=false,
 	// privacy.public_allowlist_default=true). New policy surface → ph
 	// must shift so verifiers detect the schema change.
-	goldenHashDefaults = "bfaf53a86f0a97d30975349cbe0ba09947dd42b413cd04c4076114e434ad91e8"
+	// Re-bumped on the inference-engine wiring PR: Learn now carries an
+	// inference.floors substruct (min_sessions/min_events/min_windows).
+	// Floors are detection-relevant (they decide stable vs
+	// never_confirmed at compile time), so they must flow into ph so
+	// verifiers detect deployments that loosen the exposure gates.
+	goldenHashDefaults = "2fc2087f2baaf55caeff6aea778760eee3f2e16a36687473ab29e338c95ea624"
 
 	// goldenHashRichConfig pins the hash for goldenRichYAML loaded via
 	// config.Load, post-ApplyDefaults + Validate. Covers a broad,
@@ -47,7 +52,11 @@ const (
 	// the rich fixture now carries a Learn block with enabled=false +
 	// privacy.public_allowlist_default=true defaults, so ph must shift
 	// in lockstep with the Defaults() bump above.
-	goldenHashRichConfig = "31c45f2e4131f4d780c6452192a98928d91b69a46c21e1858dfb5125d5e15dcb"
+	// Re-bumped on the inference-engine wiring PR: see goldenHashDefaults
+	// note above. Floors decode as zero in the rich fixture (which does
+	// not set them) and Resolved() supplies defaults at runtime, so the
+	// rich-config hash shifts in lockstep with Defaults().
+	goldenHashRichConfig = "19acd838ef7d36ed467077dd56a545e60808a5155fa61715aa5f889ee5f7aac7"
 )
 
 // goldenRichYAML is the canonical fixture for goldenHashRichConfig. It
--- a/internal/config/learn_test.go
+++ b/internal/config/learn_test.go
@@ -549,6 +549,9 @@ func TestLearn_YAMLRoundTrip(t *testing.T) {
 	cfg.Learn.CaptureDir = "/var/lib/pipelock/learn"
 	cfg.Learn.Privacy.SaltSource = "${PIPELOCK_REDACT_SALT}"
 	cfg.Learn.Privacy.PublicAllowlistDefault = false
+	cfg.Learn.Inference.Floors.MinSessions = 7
+	cfg.Learn.Inference.Floors.MinEvents = 30
+	cfg.Learn.Inference.Floors.MinWindows = 4
 
 	out, err := yaml.Marshal(cfg)
 	if err != nil {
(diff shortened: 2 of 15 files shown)


luckyPipewrench:
Addressed all three findings on `fa229b00`:

1. **Metrics helpers (medium):** added typed `InferenceOutcome` and `FloorFailure` string aliases with exported canonical constants (`OutcomeNeverConfirmed/Brittle/Stable`, `FloorSessions/Events/Windows`). Non-canonical values now drop silently at the helper boundary, so a caller passing a typo or drift-prone label cannot expand cardinality. Layering preserved by keeping the enum local to the metrics package; cross-package wire alignment with `inference.Confidence.String()` is asserted by a metrics-side test rather than an import.

2. **Registration probes (medium):** the two new helpers in `TestRegisterLearnMetrics_RegistersAllFour` now use canonical labels (`OutcomeStable`, `FloorSessions`). Added `TestRecordInferenceClassification_DropsNonCanonical` and `TestRecordInferenceFloorFailure_DropsNonCanonical` covering empty string, case variants, and arbitrary payloads, asserting no series leak into the registry. Pre-existing helper probes are out of scope for this PR.

3. **EnforcedValue (low):** clamped to `>= 0` so headroom below `-1` cannot produce nonsensical negative ceilings. Headroom in `[-1, 0)` still shrinks the budget per the original contract. Added `TestEnforcedValue_ClampsAtZero` with five boundary rows including the empty-budget case. Updated godoc explains the fail-closed-but-valid semantics.

100% line coverage retained on every changed function. Full repo race tests, lint, and Linux + Windows cross-builds (default + enterprise tags) all green.

DeathCamel58 (approved):

PR #454, opened 2026-04-29
feat(contract): path normalization with cardinality cap and operator pin/split

luckyPipewrench:
## Summary

Adds the `internal/contract/inference/normalize` package implementing the path-normalization layer of the contract-compile inference engine: per-bucket frequency-weighted Shannon entropy with the 5-gate collapse decision, the canonical reserved-segment blocklist, and per-host cardinality capping with `_other` tail coverage. Wires the operator-affordance commands `pipelock learn split` and `pipelock learn pin` for pre-ratification adjustments. Also folds the cardinality-trust hardening onto the two pre-existing learn helpers (`RecordObservationEvent`, `RecordRegulatedDataBlocked`) so all four learn counters share the closed-allowlist contract. Addresses the four follow-up findings on the prior PR.

## What's in the new package

- **`evidence.go`**: value types matching the rule emission shape (NormalizationEvidence, Bucket, CollapsedSegment, RetainedSegment) plus collapse and retain reason wire constants. Validate method rejects malformed instances with a wrapped sentinel.

- **`reserved.go`**: the canonical reserved-segment blocklist (admin, auth, login, oauth, token, billing, settings, users, me, orgs, sso, saml, mfa, kms, vault, secret, key, password, payment, refund, transfer, withdraw, ssh, ssl, cert, ca, auth0). IsReserved accepts an operator-supplied `extras` slice; operators may extend but cannot remove from the canonical floor.

- **`path.go`**: Canonicalize applies NFC plus Unicode lowercasing plus per-segment percent-decoding and rejects non-canonical inputs (`..`, standalone `.`, double slashes, trailing slash on non-root, control chars, percent-encoded slashes, paths over 2048 chars). Frequency-weighted Shannon entropy. Decide composer runs the 5-gate test per segment position: minimum events (default 10), minimum distinct values (default 5), entropy threshold (default 3.0 bits), reserved blocklist match, and high-risk-sibling no-merge rule. Emitted evidence is deterministic across runs (segment positions sorted ascending) so the policy hash stays stable.

- **`cardinality.go`**: per-host cap (default 1000) with `_other` tail bucketing. Top-N selection by event count with stable lexicographic tie-breaking. `ShouldBlockPromotion` engages when tail event count strictly exceeds the configured percentage of host total (default 5.0) without explicit `accept_tail`. Adversarial corpus test exercises a 100K-distinct-path host bucketing into top-1000 plus tail in roughly 0.7 seconds.

## Pen-test gates the test pack pins

- `/admin` cannot collapse: even surrounded by 1000 high-entropy random siblings at the same parent, the position retains as a literal because the reserved blocklist is a position-level decision.
- `/v1/admin` and `/v1/users` cannot merge: when both are present at the same parent and either appears in the high-risk-sibling list, the no-merge gate fires.
- 100K distinct synthetic paths bucket into top-1000 plus a single `_other` tail with the correct event count sum, and the promotion-block correctly engages because tail is roughly 99% of events.
- Non-canonical paths hard-fail at Canonicalize before reaching Decide.

## Config surface

Adds `learn.inference.normalization.{algorithm, min_events, min_distinct_values, entropy_threshold_bits, reserved_segments_extra, cardinality_cap_per_host, tail_promotion_block_pct}` to the existing `learn:` block. Algorithm is locked to `frequency_weighted_entropy_v1`; numeric fields validate to non-negative; entropy threshold is bounded to `[0, 8.0]` bits; tail percent is bounded to `[0, 100]`; the reserved-segments-extra list rejects empty entries with a precise YAML index in the error message.

The config validator deliberately duplicates the inference package's `DecideConfig.Validate()` and `CapConfig.Validate()` so operators see `learn.inference.normalization.cardinality_cap_per_host: -1: must be non-negative` instead of an API-level field name.

## CLI handlers

`pipelock learn split --candidate <path> --rule <rule_id> [--index N] [--out <path>]` rewrites a candidate contract YAML to demote one or more collapsed segments back into literal values. With `--index 0` (default) all collapsed segments for the rule are split; with a specific index only that one moves. Writes atomically via temp file plus rename. Idempotent.

`pipelock learn pin --candidate <path> --rule <rule_id> --segment <value> [--out <path>]` adds a per-rule `pinned_segments` entry for a literal value, ensuring subsequent recompile passes treat it as reserved-equivalent for that rule. Operates at the YAML node level via `gopkg.in/yaml.v3` so comments and ordering are preserved across rewrite. Idempotent.

Neither command signs or activates the candidate; that lands with the compile pipeline in a follow-up.

## Metrics fold-in

`RecordObservationEvent` and `RecordRegulatedDataBlocked` now match the closed-allowlist contract the inference helpers got in the prior PR. New typed aliases `ActionClass` (read, derive, write, delegate, authorize, spend, commit, actuate, unclassified) and `BlockReason` (field_class_regulated, root_class_regulated, explicit_block) with exported canonical constants. Non-canonical values drop silently to prevent cardinality drift on the security-relevant counters. The `unclassified` side-effect on `learnUnclassifiedActions` is preserved when reached via the canonical `ActionUnclassified` branch.

No production callers exist yet for either helper (the prior PR shipped them without runtime wiring; that lands later), so the fold-in is contained to `internal/metrics/learn.go` plus its tests.

## Prior-PR review fixes folded in

- **BudgetStats finite contract**: filters NaN and infinite values from the input window before computing statistics so the contract math layer cannot persist non-finite values into a signed envelope. SampleCount reflects post-filter count. New `TestBudgetStats_FiltersNonFinite` covers 5 boundary rows.
- **Hash drift on omitted floors**: `policySemanticView` now resolves omitted `learn.inference.floors.*` and `learn.inference.normalization.*` to their effective defaults before hashing so a YAML omitting these and a YAML setting them explicitly to the design defaults hash identically. New regression tests `TestCanonicalPolicyHash_OmittedFloorsHashAsExplicitDefaults` and `..._OmittedNormalizationHashAsExplicitDefaults` prove the invariant.
- **Wilson extreme-alpha boundary**: `invNormalCDF` clamps `p` to `(1e-15, 1 - 1e-15)` so extreme alpha values cannot drive the rational approximation toward NaN or infinity. Documented supported alpha range.
- **Source-comment OPSEC**: scrubbed internal-tracker phrasing across every reviewer-flagged file plus follow-on touched files (`budgets.go`, `floors.go`, `opportunity.go`, `wilson.go`, `schema.go`, `metrics/learn.go`, the CLI files, the new normalize files, and the canonical golden test). Replaced with product-neutral language.

## Policy-hash impact

Adding the normalization substruct plus the Resolved pre-pass shifts the canonical policy hash twice in this cycle. Both `goldenHashDefaults` and `goldenHashRichConfig` are updated with explanatory test-file comments extending the prior bumps. The Resolved pre-pass closes a verifier-drift hole: omitted YAML and explicit defaults now hash identically.

## Coverage

- 100% line coverage on every new function in the normalize package except `Entropy` (94.7%, with one intentionally uncovered defensive `IsNaN || IsInf` clamp documented inline as a defense-in-depth on the signed-policy-hash path).
- 100% on every new CLI helper top-level function (runSplit, runPin, splitCmd, pinCmd, splitNormalization, pinNormalization). A few internal yaml.Node helpers sit at 87 to 95% on defensive nil branches that yaml.v3 cannot produce on parse; documented inline.
- 100% on every new config validator and on every metrics helper (the four hardened helpers plus the two from the prior PR).
- 1300+ config-package PASS subtests, 200+ normalize-package PASS subtests, 175+ metrics-package PASS subtests, 85+ CLI-package PASS subtests.

## Cross-build

Linux and Windows builds, default and enterprise tags, all four green. The new package is pure Go math plus stdlib (`slices`, `strings`, `unicode`, `unicode/norm`, `net/url`, `errors`, `fmt`, `sort`, `math`) and `golang.org/x/text/unicode/norm` which pipelock already depends on. No syscall surface.

## Out of scope

- The full `pipelock learn compile` end-to-end pipeline (streaming JSONL ingest, deterministic recompile, signed candidate emission) lands in a later PR. This PR ships the building blocks but not the wiring that produces candidate files.
- Runtime wiring of the metrics helpers is still future work; the helpers stand ready with the closed-allowlist contract in place.
- The verifier `verify.py` is not extended; this PR does not emit signed artifacts.
- Pre-existing source-comment phrasing in `internal/contract/contract.go` and sibling files from earlier PRs is intentionally not scrubbed here; that would explode the diff into files this PR does not otherwise touch.

## Release Notes

* **New Features**
  * Added `split` and `pin` subcommands to the `learn` command for managing path segments in contract rules.
  * Introduced path normalization configuration with entropy-based bucketing and cardinality capping for improved contract inference.

* **Documentation**
  * Updated pipeline terminology from "learn-and-lock" to "contract-compile" observation pipeline throughout documentation.

* **Bug Fixes**
  * Improved handling of non-finite values in budget statistics.

### DIFF
--- a/internal/cli/learn/helpers_test.go
+++ b/internal/cli/learn/helpers_test.go
@@ -4,10 +4,23 @@
 package learn
 
 import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 
 	"gopkg.in/yaml.v3"
 )
+
+// goosWindows is the runtime.GOOS string for Windows. Extracted so the
+// symlink-skip predicate across multiple test files can share the
+// literal (goconst).
+const goosWindows = "windows"
 
 // These tests exercise low-level YAML helper functions directly so we
 // hit the defensive nil/empty branches that the canonical fixture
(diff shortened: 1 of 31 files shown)


luckyPipewrench:
Addressed all four findings on `c7e8e5fe`.

**HIGH (Filesystem trust boundary).** `loadCandidate` and `resolveOut` now reject symlinks at the directory entry level via `Lstat`, require regular files, and read with `O_RDONLY|O_NOFOLLOW` so a between-stat-and-open swap also fails closed. Per-OS shim (`nofollow_unix.go` / `nofollow_windows.go`) mirrors the existing pattern in `internal/config` so the read path is consistent across the codebase. Two new tests pin the contract: `TestLoadCandidate_RejectsSymlink`, `TestLoadCandidate_RejectsNonRegular`, `TestResolveOut_RejectsSymlinkOut`, `TestResolveOut_AcceptsNonExistentOut`.

**MEDIUM (`rebuildPathValue` segment escaping).** New `validateSegmentLiteral` rejects empty, path separators, control characters, wildcard glyphs (`*`, `?`), bracket glyphs (`[`, `]`), and overlong literals (256-byte cap). `validateRuleSegments` walks every `selector.paths[].normalization.{retained_segments, pinned_segments}[].value` on the rule before mutation runs, in both `runSplit` and `runPin`. `rebuildPathValue` now treats invalid values as wildcards as defense-in-depth so even a future caller that bypasses the entry-boundary validator cannot corrupt a path string. New tests: `TestValidateSegmentLiteral_Grammar` (16 rows) and `TestValidateRuleSegments_RejectsMaliciousYAML`.

**MEDIUM (`pin --segment` grammar).** `runPin` now applies `validateSegmentLiteral` to the trimmed segment input alongside the existing empty check. Same validator instance the YAML side uses, so CLI input and YAML-carried values share one rejection surface. Test: `TestPin_RejectsInvalidSegment`.

**LOW (audit emit).** Both runners now emit a structured JSON line on stderr alongside the human-readable stdout: `{"event":"learn_split"|"learn_pin","candidate":"...","dest":"...","rule":"...","segment":"...","index":N,"segments_changed":N,"noop":bool}`. Test: `TestEmitAuditEvent_StructuredOnStderr` parses the line back through `encoding/json` and asserts the required fields are present.

Re-verified gates: full repo race tests, lint, scan-diff, 4 cross-builds (Linux+Windows × default+enterprise), 3-layer security suite all green.

luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #467, opened 2026-05-02
feat(blockreason): X-Pipelock-Block-Reason header schema and emit package

luckyPipewrench:
## Summary

Defines the v1 vocabulary for `X-Pipelock-Block-Reason`, the response header set agents read to decide whether to retry, switch tools, or surface a useful error to the user. Lands the schema doc and the emit package now so the vocabulary can be reviewed before any block site commits to it. The transport refactor that wires every block path to call `Info.SetHeaders` follows in a separate PR.

The header set:

- `X-Pipelock-Block-Reason` (required): machine-readable code. 28 codes across egress, content, MCP, posture, and generic.
- `X-Pipelock-Block-Reason-Version` (required): schema version.
- `X-Pipelock-Block-Reason-Severity` (required): matches the existing `internal/config/schema.go` vocabulary (`info`, `warn`, `critical`).
- `X-Pipelock-Block-Reason-Retry` (required): retry hint (`none`, `transient`, `policy`).
- `X-Pipelock-Block-Reason-Layer` (optional): reuses `internal/scanner/` `Scanner*` constants verbatim so operators can correlate header-driven agent behavior with existing audit logs and Prometheus labels.
- `X-Pipelock-Block-Reason-Receipt` (optional): receipt UUID.

## Why design first

Once an agent in production reads `dlp_match`, the vocabulary is locked. Renaming a code in v2 breaks every consumer. Splitting the schema and the transport refactor into separate PRs gives reviewers a chance to challenge the vocabulary before any of the roughly 50 block sites commits to it.

## Privacy

The `Info` struct deliberately exposes only `Reason`, `Severity`, `Retry`, `Layer`, and `Receipt`. None of those fields can carry matched content, pattern names, or agent identifiers. Tests pin the surface so a future field addition trips review.

## RFC 6455 floor

`CloseFramePayload()` is bounded to RFC 6455's 123-byte close-frame limit. Optional fields drop in a documented order (receipt, layer, retry, severity, version). If even the bare `{"block_reason":"<code>"}` would overflow because the Reason value is unusually long, the helper returns a fixed sentinel rather than a malformed close frame. Test invariant pins the size ceiling across normal, truncated, overlong-reason, and floor cases.

## Anti-scope-creep

- No call sites yet. The spec status block at the top of the doc says so explicitly. Transport PR follows.
- No JSON-RPC error vocabulary for MCP stdio. Tracked separately; stdio has no HTTP layer to attach headers to.
- No agent-side retry orchestration. Pipelock emits, the agent decides.

## Release Notes

* **New Features**
  * Added block reason metadata for HTTP responses and WebSocket connections, providing detailed information about blocked requests including reason, severity, and retry guidance.

* **Documentation**
  * Added specification for `X-Pipelock-Block-Reason` header format and WebSocket close-frame behavior.

* **Tests**
  * Added comprehensive test coverage for block reason functionality.

### DIFF
--- a/docs/specs/block-reason-header.md
+++ b/docs/specs/block-reason-header.md
@@ -15,7 +15,7 @@ This document is the canonical schema. Once an agent in production reads `dlp_ma
 | `X-Pipelock-Block-Reason-Severity` | yes (on every block) | `critical` | Matches pipelock's existing severity vocabulary: `info`, `warn`, `critical`. |
 | `X-Pipelock-Block-Reason-Retry` | yes (on every block) | `none` | Retry hint: `none` (permanent), `transient` (retry with backoff), `policy` (retry only after policy change). |
 | `X-Pipelock-Block-Reason-Layer` | optional | `dlp` | Scanner pipeline layer label. Uses `internal/scanner/` `Scanner*` constants verbatim so operators can correlate header-driven agent behavior with their existing audit logs and Prometheus labels. See the layer-label vocabulary below. |
-| `X-Pipelock-Block-Reason-Receipt` | optional | `01HZ8...` | Receipt ID. Lets the agent fetch the matching receipt (via the receipt-transports endpoint) for additional context. |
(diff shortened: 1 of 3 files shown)


luckyPipewrench:
@coderabbitai reveiw

BlueAceFrost (approved):

PR #468, opened 2026-05-02
test(capture): add synthetic replay regression harness

luckyPipewrench:
## Summary

- Adds a deterministic, package-level regression harness for the learn-and-lock policy compiler. This is NOT full CLI / system replay coverage. Closes v2.4 done-state criterion 2(c).
- 3 sessions, around 12 capture entries each, across 4 surfaces (URL / DLP / response / tool_policy). Compile output, replay diff, and the rendered corpus JSONL are byte-snapshotted against checked-in goldens.
- Wires `make test-replay-harness` and a dedicated CI step ("Replay harness (deterministic regression)") that runs before the broader OSS test job so golden drift fails fast.

No production code changes. New file is `_test.go` only.

## Scope (explicit non-goals)

- Does NOT exercise the `pipelock learn compile` CLI or `readCompileInputs` path. Future CLI-level coverage would warrant its own harness.
- DLP / response surfaces are summary-only because no sidecar escrow is generated. Future sidecar-aware coverage should add a second fixture rather than stretch this one.

## Why now

The next release ships the enforce/promote phase of the policy compiler. A silent regression in compile output would corrupt every signed contract operators ratify after that point. This harness catches byte-level drift in compile / inference / signing logic before it ships.

## Refresh procedure

```
go test ./internal/capture -run TestReplayHarness -update
```

Inspect the diff manually before committing. Golden drift means a logic change altered the byte output. Provenance fields (PipelockVersion, PipelockBuildSHA, GoVersion, ModuleDigestRoot) are pinned to synthetic constants in the harness compile config so the goldens stay byte-stable across the Go-version CI matrix and across developer machines. The compile manifest build-info-derived fields (`module_digest_root`, `module_digests`) are recomputed by the emit layer regardless of the pinned config and are stabilized to placeholders before snapshot comparison.

## Summary by CodeRabbit

* **Tests**
  * Added a deterministic replay-harness test suite that validates capture, compile, and replay outputs against golden artifacts and enforces replay-diff invariants.

* **Chores**
  * Added a dedicated replay-harness test command and integrated a CI step to run the replay-harness as a deterministic regression check; included supporting golden test data for reproducible comparisons.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
Addressed all three findings from the gpt-5.4 deep review in `6e0f4940` (force-push amend).

1. (high) `harnessCandidateConfig` now configures `MCPToolPolicy` with a `system_file_deny` rule (`tool_pattern: ^write_file$`, `arg_key: ^path$`, `arg_pattern: ^/etc/`, action block). The `write_file → /etc/hosts` record now stays `block → block` (unchanged) in the replay diff. The corpus no longer normalizes a privilege-boundary weakening as expected behavior.

2. (medium) `stabilizeManifest` now validates `module_digest_root` and `module_digests` STRUCTURALLY before replacement: presence + `sha256:<hex>` shape + non-empty digest map + every value matches the prefix shape. Provenance regressions where any of those fields disappear, become malformed, or carry the wrong shape will fail CI before the placeholder swap.

3. (medium) `TestReplayHarness_ReplayDiffMatchesGolden` now asserts a security invariant: every `tool_policy` record whose `OriginalAction == block` must keep `CandidateAction == block` after replay. A future scenario that legitimately needs to test tool-policy-allow drift must land in a separate test function with its own golden.

Replay diff totals after fix: total=37, replayed=31, new_blocks=3, new_allows=3 (previously 4), unchanged=25 (previously 24), summary_only=6.

BlueAceFrost (approved):

PR #470, opened 2026-05-02
chore: SPIFFE IP rejection, blockreason pairing helpers, tombstone scope, conformance fixtures

luckyPipewrench:
## Summary

- Four targeted v2.4 pre-tag fixes from the pre-release sweep.
- `internal/envelope/spiffe.go` rejects IPv4/IPv6 literals as SPIFFE trust domains. SPIFFE-ID section 2 prohibits raw IP addresses; the prior implementation accepted them, which would let a federation partner impersonate a domain by claiming a numeric host.
- `internal/blockreason/blockreason.go` exports `SeverityFor(reason)`, `RetryFor(reason)`, `NewForReason(reason)`, and `MustNewForReason(reason)`. The exported pair-lookup helpers encode the canonical reason-to-severity-and-retry mapping from `docs/specs/block-reason-header.md` so production call sites can build an Info without tracking the spec table manually. Production callers should default to `NewForReason` so a future call site outside `internal/proxy` cannot accidentally emit a mismatched pair.
- `internal/contract/tombstone.go` scopes the `Tombstone` type doc comment from "prevents reactivation" to evidence-marker semantics. `Store.ValidateEnvelope` does not cross-check tombstones during activation today, so a tombstoned contract hash can be re-promoted under v2.4 semantics. The tombstone exists so external auditors can prove the operator declared the prior contract redacted; activation enforcement is a v2.5 follow-up. The previous comment overstated the guarantee.
- `internal/contract/receipt/golden_vectors_test.go` adds `TestGolden_EvidenceReceiptPromoteCommitted` and `TestGolden_EvidenceReceiptShadowDelta` with corresponding fixtures under `testdata/golden/`. The `pipelock-verify-python` 0.2.0 release uses these to prove byte-for-byte JCS preimage parity with the Go reference for the two load-bearing audit kinds beyond `proxy_decision`.

## Behaviour changes

- A federation peer that previously claimed a SPIFFE actor under an IP-literal trust domain (`spiffe://192.0.2.1/agent/foo`) is now rejected with `SPIFFE actor trust domain must be a DNS name, not an IP address`. Operators using IP literals must migrate to DNS names. No production deployment should be relying on this; SPIFFE-ID section 2 has prohibited it since the spec was published.
- The `Tombstone` doc comment correction is cosmetic. No behaviour change.
- The `blockreason` helper additions are additive. No existing call site changes; the existing `New` constructor is untouched. Future call sites get a recommended path that prevents pair-mismatch bugs.
- The new golden vectors are additive test data. No production code path consumes them.

### DIFF
--- a/internal/blockreason/blockreason.go
+++ b/internal/blockreason/blockreason.go
@@ -433,7 +433,10 @@ func SeverityFor(reason Reason) Severity {
 		MediaPolicy,
 		ParseError,
 		Timeout,
-		PatternUnavailable:
+		PatternUnavailable,
+		CompressedResponse,
+		BrowserShieldOversize,
+		BlockReasonOverflow:
 		return SeverityWarn
 	default:
 		return SeverityCritical
@@ -453,7 +456,10 @@ func RetryFor(reason Reason) Retry {
 		EscalationLevel,
 		RedactionFailure,
 		Timeout,
-		PatternUnavailable:
+		PatternUnavailable,
+		OutboundEnvelopeFailed,
+		SessionAnomaly,
+		BlockReasonOverflow:
 		return RetryTransient
 	case DomainBlocklist,
 		PathEntropy,
@@ -464,7 +470,9 @@ func RetryFor(reason Reason) Retry {
 		ToolPolicyDeny,
 		SessionBinding,
 		AuthorityMismatch,
-		NotEnabled:
+		NotEnabled,
+		CompressedResponse,
+		BrowserShieldOversize:
 		return RetryPolicy
 	default:
 		return RetryNone
--- a/internal/blockreason/blockreason.go
+++ b/internal/blockreason/blockreason.go
@@ -413,3 +413,85 @@ func mustMarshal(p closeFramePayload) string {
 	}
 	return string(b)
 }
+
+// SeverityFor returns the canonical severity for a block-reason code per
+// docs/specs/block-reason-header.md. The mapping is locked at v1; new
+// reasons must add a case here at the same time the constant is added.
+// Unknown reasons return SeverityCritical so a missing-mapping bug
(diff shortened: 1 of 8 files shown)


luckyPipewrench:
**On the `/review deep` findings**

**Finding 1 (tombstone, high):** the diff in this PR scoped the `Tombstone` type comment to evidence-marker semantics. It did not introduce the enforcement gap; it stopped overstating it. Activation-time tombstone cross-checking is a real feature (store-layer integration, threat-model review, transport-parity tests, red-team coverage) and lands in v2.5. v2.4 ships with accurate docs about current behavior; v2.5 adds enforcement. Tracked as a v2.5 follow-up.

**Finding 2 (blockreason, medium):** the concrete factual half is fixed in `eab38488`. `SeverityFor` and `RetryFor` diverged from `docs/specs/block-reason-header.md` for five reasons: `compressed_response`, `browser_shield_oversize`, `outbound_envelope_failed`, `session_anomaly`, and `block_reason_overflow`. The new `TestSpecCanonicalPairs_VocabularyCoverage` guard fails the build if a Reason is added without a spec entry (or vice versa), so future drift between the markdown spec and the helper functions cannot pass CI.

Migrating the three production callers in `internal/proxy/blockheaders.go` and `internal/mcp/proxy_http.go` to `NewForReason` (eliminating the duplicate local mapping in `proxy/blockheaders.go`'s `severityFromReason` and `retryFromReason`) is the architectural fix for "production code can emit non-canonical pairs". After this commit's spec alignment, that migration becomes a behavior-preserving refactor. Deferred to v2.5 to keep v2.4 pre-tag PR scope tight; tracked alongside the tombstone follow-up.

BlueAceFrost (approved):

PR #474, opened 2026-05-03
fix(capture,config): SessionID empty + caller-side metadata + race + e2e regression test

luckyPipewrench:
## Summary

Fixes the capture pipeline so learn-and-lock observation produces usable events end-to-end. The bug was structural: every `Observe*` call site in proxy and MCP packages constructed verdict records without setting `SessionID`, the writer worker rejected each entry as "empty session ID" and only drop sentinels reached disk. `learn compile` emitted zero rules from any traffic; `learn shadow` had nothing to evaluate.

## Commits

`3c909452` populates `SessionID` at every `Observe*` call site. HTTP proxy paths use `CeeSessionKey(agent, clientIP)`; MCP paths use `"mcp-" + opts.Transport`. Defensive `ActionAllow` default in the writer for empty `effectiveAction` so shadow can compute a non-empty `original_verdict`.

`1e732782` does caller-side stamping of `EffectiveAction`, `ConfigHash`, and `Profile` at every call site. Intercept and reverse response capture moved to fire AFTER suppression filtering (previously recorded findings the runtime later suppressed). `learn compile` and `learn shadow` now walk one level deeper into agent|ip session subdirs so contracts find their input automatically. `MCPProxyOpts.ConfigHashFn` closure for hot-reload listener freshness. Plus a `CanonicalPolicyHash()` cache race fix found by `-race`: atomic.Pointer plus sync.Once replaces the unsynchronized cache write. The race predates this work but the new high-frequency capture read path made it visible.

`ff082ab0` adds an end-to-end regression test that drives a fetch through `proxy.handleFetch` with a real `capture.Writer` attached, drains the writer, walks the session subdir, opens the recorder JSONL, and asserts `session_id`, `effective_action`, `config_hash`, `profile`, and `agent` all populate. Locks the fix in.

## Behavior

After the fix, capture entries land in real session subdirs with non-empty `session_id`, `effective_action`, `config_hash`, `profile`, and `agent`. `learn compile` ingests the captures and emits rules; `learn ratify` accepts them to enforce; `learn shadow` replays records and computes deltas; `learn promote` signs both intent and committed receipts. The defensive writer `ActionAllow` fallback stays as a backstop invariant for any future call site that misses caller-side stamping.

## Out of scope

- `ActionClass` population. The classifier wiring is documented as next-phase work; review markdown will continue to show 100% unclassified until that lands.
- Promote staging flow (manual `cp` into history dir vs a `learn stage` command). Separate workflow polish item, not a capture-pipeline unblocker.

## Summary by CodeRabbit

* **New Features**
  * Added session filtering controls for selective replay of captured sessions
  * Enriched capture metadata with configuration hash, profile, and original session identifiers for improved audit trails

* **Bug Fixes**
  * Verdict actions now explicitly default to "allow" for clean scan results instead of empty values
  * Enhanced session directory validation for safer capture handling and identity tracking

### DIFF
--- a/internal/capture/loader.go
+++ b/internal/capture/loader.go
@@ -32,6 +32,8 @@ type ReplayOptions struct {
 	EscrowPrivateKey []byte
 	// Contract enables contract-aware replay for supported surfaces.
 	Contract *contract.Contract
+	// SessionFilter optionally limits replay to matching session directory names.
+	SessionFilter func(string) bool
 }
 
 type replayEscrowKey struct {
@@ -98,9 +100,13 @@ func LoadAndReplayWithOptions(cfg *config.Config, sessionsDir string, opts Repla
 	// sessions are processed in alphabetical order by session ID.
 	var sessionNames []string
 	for _, de := range dirEntries {
-		if de.IsDir() && de.Name() != metaSessionID {
-			sessionNames = append(sessionNames, de.Name())
+		if !de.IsDir() || de.Name() == metaSessionID {
+			continue
 		}
+		if opts.SessionFilter != nil && !opts.SessionFilter(de.Name()) {
+			continue
+		}
+		sessionNames = append(sessionNames, de.Name())
 	}
 	sort.Strings(sessionNames)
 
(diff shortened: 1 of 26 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #478, opened 2026-05-04
chore: prepare v2.4.0 release

luckyPipewrench:
## Summary

Release prep for v2.4.0. Documents every PR since v2.3.0, adds four new operator guides (block-reason header, federation, learn-and-lock, health endpoint), and ships four corrections caught by independent review.

v2.4 ships the full live-lock arc end-to-end: compile, candidate signing, shadow replay, ratification, signed active-manifest workflow, AND live proxy enforcement of promoted contracts on every URL-bearing transport plus the MCP tool-call surface. The runtime is wired into the proxy request path on forward (absolute-URI and CONNECT), reverse and redirect-refresh chains, intercept, `/fetch`, WebSocket handshake, MCP HTTP listener and stdio-to-HTTP bridge, MCP stdio subprocess wrap, and the `mcp_tool_call` rule kind via `runtime.EvaluateMCP`. Scanner block always wins over contract allow on every gated transport, proven by named regression tests per surface plus a unified decision matrix harness.

## Doc accuracy fixes

Block-reason header: replaces "every block path" with "every HTTP-capable block path", documents the production-path matrix exemptions (`tool_poisoning`, `tool_chain_blocked`) where MCP-internal blocks fire at the JSON-RPC layer with no HTTP response surface, expands the reason vocabulary table with seven previously undocumented codes (`outbound_envelope_failed`, `redirect_scan_denied`, `session_anomaly`, `cross_request_deny`, `compressed_response`, `browser_shield_oversize`, `block_reason_overflow`), and reframes the optional `X-Pipelock-Block-Reason-Receipt` as reserved (no v2.4 production emit site).

Federation: adds the SPIFFE IPv4/IPv6 trust-domain rejection, the canonical-path enforcement on the workload, the `pipelock_envelope_verify_total{result}` observability surface (`disabled` / `verified` / `missing` / `failed`), and the `inbound_verify_*` failure pattern table (`signature`, `digest`, `parse`, `not_trusted`, `replay`, `expired`, `missing`, `failed`) replacing the old free-form event names. Adds an explicit caveat that `X-Pipelock-Agent` is caller-asserted and must not be trusted as an authenticated identity source.

Learn-and-lock: adds the capture metadata hardening section (`session_id_original` semantics, `validateCaptureSessionDir` poisoning defence) and the operator soak counters with the actual closed sanitisation-reason label set (`unsafe_path` / `overlength` / `unknown`). Operator-runbook subsections cover the trust-roster bootstrap recipe using the public CLI commands shipped in v2.4 (`pipelock signing key generate`, `pipelock signing roster build`), the receipt-signing key requirement so promote-time receipts validate against the roster, the contract-store history staging step before `learn promote`, and the shuttle-pod pattern for shipping `active.json` into a `FROM scratch` pipelock pod's PVC.

Health endpoint: clarifies the `subsystems` map is gated on `expose_subsystems: true` (default `false`); the HTTP 503 + status string still surface wedges regardless of whether the per-subsystem breakdown is exposed.

## --header hardening detail

Expanded the v2.4 `pipelock mcp proxy --header` entry to specify the actual validation surface: header names must be RFC 7230 tokens, values reject ASCII control bytes, DEL, CRLF, and Unicode whitespace before trimming, and six headers (`Content-Type`, `Accept`, `Mcp-Session-Id`, `Content-Length`, `Transfer-Encoding`, `Host`) are blocked case-insensitively at both the CLI flag-parser and the HTTP transport-level guard.

## Compliance maps

`docs/compliance/owasp-mcp-top10.md`, `docs/compliance/nist-800-53.md`, and `docs/compliance/eu-ai-act-mapping.md` align with the new HTTP-capable / JSON-RPC-error-metadata distinction. The eu-ai-act "Resilient against unauthorized alteration" row reframes from Full to Partial: the binary scans every request that reaches it, but the topology that forces traffic through Pipelock is deployment-enforced, not binary-enforced. The "Code execution sandboxing" row reframes from "no coverage" to partial OS-dependent coverage.

## Conformance notes

Doc claims were cross-checked against shipped code at the rebased branch HEAD. The `internal/contract/runtime` package IS wired into the proxy request path through `internal/proxy/contractgate.go` and `internal/mcp/contractgate.go`, both of which construct loaders that pass the deployment-environment tuple, the pinned roster fingerprint, and the runtime mode through to `runtime.NewLoader`. The `health_watchdog.expose_subsystems` gate is `cfg.HealthWatchdog.ExposeSubsystems` at `internal/proxy/proxy.go`. The federation failure-pattern names are sourced from `internal/proxy/proxy.go`. The `WithReceipt` builder has zero production callers in the current tree. CI runs the full `go test -race -count=1 ./...` and `golangci-lint run ./...` matrix.

## Summary by CodeRabbit

* **New Features**
  * v2.4: Learn-and-Lock behavioral contracts, EvidenceReceipt v2 support, federation with inbound verification + replay protection, SPIFFE actor/key-directory support, standardized X-Pipelock-Block-Reason header, redaction provider plugin shape, enhanced SSE injection detection, health-watchdog wedge detection.

* **Documentation**
  * New/expanded guides: Learn-and-Lock, Federation, Block-Reason header, Health endpoint, SSE streaming, transport modes, deployment recipes, compliance mappings, README updates.

* **Tests**
  * Added/updated tests for SSE canonicalization, joined-payload rescan, recorder behavior.

* **Chores**
  * CHANGELOG entry, chart appVersion bump to v2.4.0, Makefile release-check update, license/charter/security docs and pages workflow removal.


LICENSE
```diff
@@ -1,3 +1,4 @@
+
                                  Apache License
```
DeathCamel58, line 2: Why the license change? No longer Apache 2.0 when you change the document.

DeathCamel58 (changes requested):

DeathCamel58 (dismissed):

LICENSE
luckyPipewrench: This was intentional: the copy on main had drifted from the canonical Apache-2.0 text in a few wording spots and was missing the standard appendix. This restores the standard Apache-2.0 text; the only project-specific substitution is the appendix copyright line for `Copyright 2026 Joshua Waldrep`

DeathCamel58 (approved):

PR #525, opened 2026-05-14
fix(claude-hook): close fail-open path for unknown tools

gabalint:
## Summary

The Claude Code `PreToolUse` hook had a fail-open path that contradicts the project's "fail-closed everywhere" hard rule:

- The installed matchers were `Bash|WebFetch|Write|Edit` and `mcp__.*`. Every other tool (`WebSearch`, `Task`, `NotebookEdit`, and any future tool Anthropic ships) never reached pipelock at all.
- For tool names that *did* reach the hook but were not parsed specifically, `claudePayloadToAction` returned `(nil, nil)` with the comment `// Unknown tool: default allow`, which `runClaudeHook` treated as approve.

Net effect: an agent could call `WebSearch` with an Anthropic API key in the query, or any future tool with secrets in its arguments, and pipelock would silently approve it.

This PR closes that path:

- Replaces the two split matchers with a single `.*` matcher so every tool call reaches the hook.
- Adds a new generic `EventToolUse` kind in `internal/decide` with a `decideToolUse` handler. It extracts every string from `tool_input` via the existing `ExtractAllStringsFromJSON` helper and runs DLP + (optional) injection scanning on the joined text.
- Malformed `tool_input` is now block-level evidence rather than a silent allow.
- `claudePayloadToAction` falls through to this catch-all in its `default` arm, so the specific Bash/WebFetch/Write/Edit/MCP paths still get their tool-aware scanning; everything else gets the generic DLP+injection sweep.

No new direct dependencies. No config-schema changes. No public CLI changes (hook entries are regenerated by `pipelock claude install`; users will see one `.*` group instead of two split groups after re-running it).

## Why

- `CLAUDE.md` Hard Rules: *"Never bypass fail-closed defaults. ... If in doubt, block."*
- `CLAUDE.md` Security Invariants: *"Allowlist/suppression must not bypass content scanning... unless the exception is deliberate, documented, and tested."* The old `nil, nil` path was none of those.
- Transport parity: the forward proxy, fetch proxy, WebSocket proxy, and MCP proxy all scan unknown traffic. The Claude Code hook was the odd one out.

## Test plan

- [x] `go test -race -count=1 ./internal/cli/setup/... ./internal/decide/...` (only unrelated `TestVscodeInstall_WithRelativeConfigPersistsAbsolutePath` fails — pre-existing macOS `/var` vs `/private/var` symlink artifact on `main`, untouched by this PR).
- [x] `go vet ./internal/cli/setup/... ./internal/decide/...` clean.
- [x] `gofumpt -l` clean on the four touched files.
- [ ] Maintainer: verify against full CI matrix (Go 1.25 + 1.26, golangci-lint v2, govulncheck, CodeQL, self-scan).
- [ ] Manual: run `pipelock claude install` against a fresh `~/.claude/settings.json`, confirm a single `.*` matcher group is written; re-run and confirm idempotence.
- [ ] Manual: invoke `WebSearch` with a secret in the query and confirm the hook denies.

New tests in this PR:
- `TestClaudeHookCmd_UnknownTool_CleanInputAllows` — sanity: unknown tool with clean input still allows.
- `TestClaudeHookCmd_UnknownTool_SecretInArgsDenies` — secret in arbitrary tool's args is now caught.
- `TestClaudeHookCmd_WebSearch_SecretInQueryDenies` — concrete regression for the `WebSearch` exfil path.
- `TestDecide_ToolUse` — 5 subcases: clean, secret, malformed JSON (fail-closed), empty input, nested secret in object.
- `TestDecide_NilPayload` extended to cover `EventToolUse`.
- `TestMergeClaudeHooks_Fresh` / `_Idempotent` / `TestClaudeSetupCmd_Idempotent` updated to expect one matcher group instead of two.

## Compatibility

Users who already ran `pipelock claude install` keep their existing two-group config and the hook still works (both matchers route to the same binary, which now handles the unknown-tool case correctly). Re-running install will collapse the two groups into the single `.*` group via the existing remove-then-add merge logic.

## Summary by CodeRabbit

* **Security Improvements**
  * Enhanced Claude Code tool integration with safer, more consistent event routing for tool calls.
  * Unknown and future tools now undergo security scanning for secrets and injection attempts rather than being silently allowed.
  * Tool inputs are scanned for data leakage and injection vulnerabilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luckyPipewrench/pipelock/pull/525)

### DIFF
--- a/internal/cli/setup/claude.go
+++ b/internal/cli/setup/claude.go
@@ -190,10 +190,6 @@ func runClaudeHook(cmd *cobra.Command, configFile string, exitCodeMode bool) (re
 		return claudeResult(cmd, exitCodeMode, payload.HookEventName, decisionDeny,
 			"pipelock: "+err.Error())
 	}
-	if action == nil {
-		// Unknown tool: default allow.
-		return claudeResult(cmd, exitCodeMode, payload.HookEventName, decisionAllow, "")
-	}
 
 	// Decide.
 	decision := decide.Decide(cmd.Context(), cfg, sc, pc, *action)
@@ -231,7 +227,10 @@ func claudeResult(cmd *cobra.Command, exitCodeMode bool, hookEventName, permissi
 }
 
 // claudePayloadToAction routes a Claude Code tool_name to a decide.Action.
-// Returns nil action for unknown tools (default allow).
+// Known built-in tools (Bash, WebFetch, Write, Edit) and MCP tools route to
+// their tool-aware scanning pipelines. Everything else falls through to a
+// generic catch-all that runs DLP + injection on every string in tool_input,
+// so unknown or future tools cannot silently exfiltrate secrets.
 // Returns error for known tools with unparseable tool_input (fail-closed).
 func claudePayloadToAction(p claudeCodePayload) (*decide.Action, error) {
 	action := decide.Action{Source: "claude-code"}
@@ -298,8 +297,15 @@ func claudePayloadToAction(p claudeCodePayload) (*decide.Action, error) {
 		return &action, nil
 
 	default:
-		// Unknown tool: return nil to signal default allow.
-		return nil, nil //nolint:nilnil // nil,nil signals "unknown tool, allow by default"
+		// Generic catch-all: scan every string in tool_input for DLP +
+		// injection. Closes the fail-open path for tools we don't parse
+		// specifically (WebSearch, Task, NotebookEdit, future tools, etc.).
+		action.Kind = decide.EventToolUse
+		action.ToolUse = &decide.ToolUsePayload{
+			ToolName:  p.ToolName,
+			ToolInput: string(p.ToolInput),
+		}
+		return &action, nil
 	}
 }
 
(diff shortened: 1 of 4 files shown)


luckyPipewrench (commented):
Looks great! If you'll sign the [Contributor License Agreement](https://cla-assistant.io/luckyPipewrench/pipelock?pullRequest=525) CI can run and get this one moving. Thanks!

luckyPipewrench:
Thanks for the report and patch! This fail-open class is now fixed on `main` through the v2.5 hardening work in #568, with the implementation rewritten there rather than merging this branch directly due to the CLA left unsigned.

Current `main` routes unknown Claude Code tool names through the generic tool-use scanner with the full `tool_name` and `tool_input`, denies secret-bearing unknown-tool payloads, allows clean unknown-tool payloads, and returns an explicit error for null `tool_input`. The release docs in #529 now document that behavior for v2.5.

Closing this as superseded by the merged implementation.

PR #534, opened 2026-05-16
feat(rules): add opt-in ATR rules adapter

eeee2345:
## What

Adds an opt-in adapter at `internal/rules/atrimport/` that loads
Agent Threat Rules (ATR) YAML files and converts the regex-based
subset into the existing `config.DLPPattern` and
`config.ResponseScanPattern` types.

The adapter does not touch the signed bundle pipeline at
`internal/rules/`. Operators who do not call `atrimport.LoadDir`
see no behaviour change.

## Why

ATR is an MIT-licensed external detection standard for AI agent
threats (419 rules across 10 categories as of 2026-05). It is
already in production at:

- microsoft/agent-governance-toolkit (PR #908, merged)
- cisco-ai-defense/skill-scanner (PR #79, merged)
- MISP taxonomies and galaxy (#323, #1207, merged)
- OWASP Agentic-Security-Resources-Hub (#74, merged)

Operators running Pipelock against agents that also need to satisfy
those ecosystems were previously copy-pasting regexes by hand. This
adapter makes the import a one-line call.

As discussed on open-telemetry/semantic-conventions-genai#132, the
two projects share an interest in standard names for agent-threat
detection attributes; this PR is the read-only feeder side of that
conversation, not the OTel side.

## Design

- Read-only. ATR's release cadence and trust root are separate from
  Pipelock's signed bundles, so wrapping ATR into the signed bundle
  format would conflate two trust models. The adapter keeps them
  apart.
- Surfaces what is dropped. ATR rules with `detection_tier:
  behavioral`, multi-condition `all` rules, or non-regex operators
  are recorded in `Result.Skipped` with a per-rule reason, not
  silently ignored.
- Safety caps: 1 MiB per-file size limit, 4096-char regex ceiling,
  2000-rule cap per call.
- Filters: `MinSeverity`, `IncludeExperimental`, `ScanTargets`
  allow list. All default to the conservative posture.

## Files

- `internal/rules/atrimport/atrimport.go` (241 LOC)
- `internal/rules/atrimport/atrimport_test.go` (181 LOC, 7 test cases)
- 5 testdata YAML fixtures
- `docs/guides/atr-rules-import.md`

## License

Adapter code is Apache-2.0, matching the rest of the core. ATR
upstream is MIT licensed. Both are compatible.

---

**Local CI status:** code self-reviewed against the agent's design; Go is not installed on the authoring machine, so go test/go build were not run locally before push. Pipelock CI should validate. Happy to iterate on any test failures or design concerns.

### DIFF
--- a/docs/guides/atr-rules-import.md
+++ b/docs/guides/atr-rules-import.md
@@ -1,0 +1,68 @@
+# Importing Agent Threat Rules (ATR) into Pipelock
+
+This guide documents the optional `atrimport` adapter at
+`internal/rules/atrimport/`. It loads YAML rules from the external
+Agent Threat Rules (ATR) project and converts the regex-based subset
+into Pipelock pattern types.
+
+ATR is an MIT-licensed community-maintained detection standard for AI
+agent threats. Upstream: https://github.com/Agent-Threat-Rule/agent-threat-rules.
+
+The adapter is read-only and does not touch the signed bundle pipeline.
+Operators who do not call `atrimport.LoadDir` will see no behaviour
+change.
+
+## Scope and limits
+
+The adapter intentionally imports only the subset of ATR that maps
+cleanly onto Pipelock's existing pattern types:
+
+- Rules with `detection_tier: pattern` (regex-based).
+- Conditions with `operator: regex`.
+- Single-condition rules, or multi-condition rules with
+  `condition: any` (logical OR).
+
+Rules that fall outside that subset are recorded in
+`Result.Skipped` with a per-rule reason, including:
+
+- `detection_tier: behavioral` and other non-pattern tiers.
+- `condition: all` with multiple conditions (requires boolean
+  composition that the scanner does not model today).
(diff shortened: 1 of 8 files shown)


luckyPipewrench:
Thanks for the contribution. I'm closing this draft without merge.

Pipelock's detection pipeline runs through the signed rules bundle at `internal/rules/`. Every rule is content-addressed and the bundle is fetched, verified, and hot-reloaded as a unit. An adapter at `internal/rules/atrimport/` that loads YAML and constructs DLP/response patterns at runtime bypasses that signing path, which adds a second trust model to the binary that we deliberately don't have.

If operators want to use ATR's regex subset with Pipelock, the right shape is an external converter that emits a `pipelock-rules` bundle YAML and ships through the signed bundle channel. See the `pipelock-rules` repo and `internal/rules/` for the bundle format. That keeps ATR's release cadence and trust root separate from Pipelock's (which the PR description called out as the design goal anyway), and it solves the same operator problem without parallel trust paths.

Closing here.

PR #614, opened 2026-05-25
feat(hermes): Ed25519-signed integrity check for identity anchors

luckyPipewrench:
## Summary

Adds `pipelock hermes anchors seal` and `pipelock hermes anchors verify`, an integrity check for Hermes' seeded identity-anchor files (SOUL.md, IDENTITY.md, AGENTS.md). Hermes seeds these once and never re-seeds, so an attacker who can write the agent's persistent workspace can overwrite the agent's identity and have it persist across restarts.

- `seal` records SHA-256 of each anchor and signs the manifest with an Ed25519 private key.
- `verify` (run at startup, e.g. a Kubernetes init container) recomputes the digests, checks the signature against the public key, and exits non-zero on any drifted, missing, or unreadable anchor, refusing to start the agent.

## Why a signature, not a keyed MAC

Verify needs only the public key, which can sit in-cluster with no risk; the private signing key stays off-cluster and is used only at seal time. A compromised agent or cluster therefore cannot forge a baseline. This reuses pipelock's existing Ed25519 signing primitives (`internal/signing`).

## Hardening

The signature covers the whole `{version, digest_algorithm, digests}` manifest, so an attacker who can write the baseline cannot drop, add, or alter an entry, downgrade the version, or swap in a baseline signed by another key. Verify also requires the baseline to cover exactly the requested anchor set as defense in depth. Anchor reads reject symlinks and oversized files so a tampered anchor cannot turn verify into a device read or an OOM. Anchor names and baseline keys are validated as clean relative paths (no traversal). The private signing key is permission-gated and rejected if it lives under the workspace, as is a path-based public key. Fail-closed throughout.

## Scope

Verify is a point-in-time startup check: it detects tampering that persisted from before boot and refuses to start, but it does not prevent writes while the agent runs. Closing that window is a deployment concern (mount the anchor files read-only), handled by the separate cluster identity and topology work. `MEMORY.md` is intentionally not an anchor (it changes legitimately as the agent curates memory).

## Summary by CodeRabbit

* **New Features**
  * Added `hermes anchors seal` command to create integrity baselines for anchor files using cryptographic signatures.
  * Added `hermes anchors verify` command to validate anchor files against baselines and detect tampering or missing files.
  * Supports custom anchor sets, workspace paths, and JSON-formatted verification reports.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luckyPipewrench/pipelock/pull/614?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/cli/hermes/cmd.go
+++ b/internal/cli/hermes/cmd.go
@@ -40,5 +40,6 @@ The 'hook' subcommand is the per-event subprocess entrypoint; 'install',
 	cmd.AddCommand(verifyCmd())
 	cmd.AddCommand(rollbackCmd())
 	cmd.AddCommand(hookCmd())
+	cmd.AddCommand(anchorsCmd())
 	return cmd
 }
--- a/internal/cli/hermes/integrity.go
+++ b/internal/cli/hermes/integrity.go
@@ -1,0 +1,542 @@
+// Copyright 2026 Josh Waldrep
+// SPDX-License-Identifier: Apache-2.0
+
+package hermes
+
+import (
+	"crypto/ed25519"
+	"crypto/sha256"
+	"crypto/subtle"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/luckyPipewrench/pipelock/internal/signing"
+)
+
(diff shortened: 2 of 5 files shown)


luckyPipewrench:
Closing: this anchor-integrity command is deployment-specific hardening for agents that keep persistent identity files on writable storage, not a general pipelock integration feature. The MCP-wrap integration ships in #610, and the plugin/hook coverage is already merged. The signed-baseline work stays on this branch in case it is wanted later as deployment tooling.

PR #665, opened 2026-06-03
ci: run python verifier from source to stop recurring Scorecard pin alert

luckyPipewrench:
OpenSSF Scorecard's Pinned-Dependencies check (code-scanning alert #58, and #57 before it) keeps re-firing on the verifiers workflow. The conformance gate installed the local Python verifier with `pip install --no-deps --no-build-isolation .`, and Scorecard flags any `pip install` without `--require-hashes` as unpinned. A local source install (`.`) cannot be hash-pinned, so the alert is unwinnable as written: the previous `--no-deps` plus an explanatory comment did not satisfy the heuristic because it flags the command, not the actual risk.

This stops running `pip install .` at all. The verifier package is pure-Python with a hardcoded `__version__` and a `__main__`, so the gate now runs it from its pinned checkout via `PYTHONPATH` and installs only the hash-pinned dependencies. No unpinned pip command remains in the workflow, so Scorecard has nothing to flag and the alert cannot recur for this verifier.

## Conformance notes

The conformance gate runs only in CI (it needs the external verifier checkouts and the corpus), so the Verifiers workflow run on this PR is what confirms run-from-source resolves `python -m pipelock_verify` correctly.

## Summary by CodeRabbit

* **Chores**
  * Updated dependency handling in the conformance verification workflow to install only hash-pinned runtime requirements and isolate verifier imports, improving reproducibility and ensuring the verifier is loaded from the checked-out source.

[Review Change Stack](https://app.coderabbit.ai/change-stack/luckyPipewrench/pipelock/pull/665?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #751, opened 2026-06-13
feat(doctor): flag inert exemptions and semantic config mismatches

luckyPipewrench:
`pipelock doctor` gains `config`-surface checks for exemptions that parse cleanly but no enabled scanner can honor, the failure mode that makes an operator think a false positive is fixed while the block silently persists.

- Warns when: a `suppress:` entry names a pattern that matches no active DLP or response-scanning pattern (inert); a `suppress:` entry names a response-scanning pattern while `response_scanning` is disabled; a `suppress:` entry names a DLP pattern while no suppress-consulting proxy scanner (`request_body_scanning`, SSE) is enabled, with the correct remediation being `dlp.patterns[].exempt_domains` because URL-query DLP does not consult `suppress:`; and `exempt_domains` set on a disabled scanner.
- Each finding names the correct knob in its `next` action; surfaces in text, `--json`, the summary, and the exit code.
- Conservative: a warning is only emitted when inertness is provable from the loaded config. Response scanning uses a separate pattern namespace, so it does not make a DLP-namespace suppress effective.
- Docs and tests included.

## Summary by CodeRabbit

* **New Features**
  * `pipelock doctor` now detects inert suppression entries that don't affect any enabled scanner
  * Added semantic validation for exemption configuration mismatches
  * New warnings appear in text and JSON output, and are counted in doctor summary and exit code


luckyPipewrench:
@coderabbitai review

DeathCamel58 (dismissed):

DeathCamel58 (approved):

PR #759, opened 2026-06-14
fix(conductor): de-conflate misleading publish conflict error

luckyPipewrench:
## Problem

A forward policy-bundle publish after a rollback can fail for three distinct reasons, but `pipelock conductor publish` reported all three as a single misleading "policy bundle version is stale". During a live recovery this cost a failed publish: the head had been rolled back to `vN` while `vN+1..vM` still existed, so the forward publish needed a version above the stream max `M`, not merely above the current head `N`, but the error said only "stale", and a `previous_bundle_hash` mismatch reported the same way.

## Change

The control plane now emits a machine-readable `code` in the `409 Conflict` body, and the CLI maps each to a distinct, actionable error:

- rollback attempt: the version is at or below the rolled-back head, and a publish cannot roll back (use the rollback authorization flow).
- version below stream max: publish a version above the stream's highest-ever version, not merely above the current head.
- previous_bundle_hash mismatch: the supplied hash does not match the current stream head hash.

An older control plane that emits no `code` falls back to a generic conflict, so the change is backward and forward compatible. HTTP status semantics are unchanged (still 409); only the error classification and message are clarified. There is no change to which publishes succeed or fail.

Free-tier operability.

## Changelog (for the pre-tag sweep, not added to CHANGELOG.md here)

Conductor: de-conflated the policy-bundle publish `409 Conflict` error. A forward publish that fails after a rollback now reports the specific cause (a rollback attempt, a version at or below the stream's highest-ever version, or a `previous_bundle_hash` mismatch) instead of collapsing all three into a misleading "version is stale".

## Summary by CodeRabbit

* **Documentation**
  * Added operator runbook guidance for policy bundle publication conflicts, including three distinct conflict scenarios and recovery steps.

* **New Features**
  * Policy bundle publication now returns machine-readable conflict codes in error responses for improved error identification and automated handling.

* **Bug Fixes**
  * Enhanced error messages for policy publication failures to distinguish between rollback attempts, version constraints, and hash mismatches.

### DIFF
--- a/docs/guides/conductor-operator-runbook.md
+++ b/docs/guides/conductor-operator-runbook.md
@@ -248,6 +248,18 @@ The server authorizes the request with the publisher token file configured on `c
 
 This runbook does not include a one-line policy signing helper because none exists in the verified CLI help. Use the signed bundle producer in your operator workflow, then publish through the API above.
 
+### Publish conflicts (HTTP 409)
+
+A forward publish can be rejected with `409 Conflict` for three operationally distinct reasons. The control plane carries a machine-readable `code` in the JSON error body so the publisher can tell them apart instead of treating every conflict as a stale version:
+
+| `code` | Meaning | What to do |
+| --- | --- | --- |
+| `rollback_attempt` | The supplied `version` is below the current (rolled-back) stream head. A publish cannot roll back. | Use the rollback authorization flow, not a publish. |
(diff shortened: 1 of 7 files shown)


luckyPipewrench:
Consolidated into #758. These two changes are the same cohesive surface (conductor Phase-A operator robustness), so they belong in one PR rather than two small ones. #758 now carries both the stream-observability endpoint/CLI and the publish-conflict error de-conflation, plus the two CodeRabbit fixes raised here (the `--version` flag wording and the rollback-attempt distinctness assertion).

PR #766, opened 2026-06-14
test(cli): make read-only-dir/config tests portable on Windows

annawhooo:
## What

Fixes five `internal/cli` tests that fail on Windows checkouts. Two distinct root causes, handled per-test rather than with a blanket platform skip.

### Fixed (Windows coverage retained)

A `t.TempDir()` path was interpolated into a **double-quoted YAML scalar**, so Windows backslashes were parsed as YAML escape sequences. A temp path segment like `...\001\test.yaml` produces `\U`/`\0`, which YAML reads as a (malformed) unicode escape:

```
parsing config ...\test.yaml: yaml: line 11: did not find expected hexdecimal number
```

Wrapping the path in `filepath.ToSlash` before interpolation keeps the config valid. The app opens forward-slash paths fine on Windows, and the later `os.ReadFile` still uses the native path variable, so nothing else changes.

- `TestRunCmd_Integration`
- `TestRunCmd_InvalidMode`
- `TestRunCmd_ReloadRejectsForwardProxyEnable`

### Skipped on Windows

`os.Chmod(dir, 0o500)` does not make a directory non-writable on Windows (POSIX mode bits don't gate directory writes), so the expected write error never occurs and the test's `expected error writing to read-only directory` assertion fails. There's no portable equivalent; the write-error path stays covered on Unix CI. Same rationale as #711.

- `TestGenerateCmd_WriteError`
- `TestGenerateDockerComposeCmd_WriteError`

## Note

stdlib `runtime` is imported as `stdruntime` because this file already imports `internal/cli/runtime` as `runtime`. No file in the tree imports both, so there was no existing alias to match; `stdruntime` is the minimal-churn choice.

## Verification

- Before: all 5 FAIL on `windows/amd64`. After: 3 PASS, 2 SKIP.
- `go test ./internal/cli/` full package: `ok`.
- `go vet ./internal/cli/`: clean.
- `-race` and `golangci-lint`/`gofumpt` not run locally (no gcc / tools not installed on the Windows dev box); they run in CI on Linux. Import placement and skip idiom verified by hand against existing conventions.

Follow-up planned: same treatment for the `internal/certgen` read-only-dir tests as a separate PR.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform reliability by standardizing how audit-log file paths are represented in generated YAML config (including initial and hot-reloaded scenarios).
  * Updated CLI “write error” test coverage to use a safe, existing output directory, ensuring consistent failures and stable error-message assertions across operating systems.


luckyPipewrench (changes requested):
One small request before merge: for the two write-error tests, could you keep Windows coverage instead of skipping by using a portable write failure? For example, set the output path to an existing directory and assert the command returns the wrapped write error. `os.WriteFile` should fail on all platforms for that case, so we do not need to rely on Unix-only `chmod 0500` behavior.

luckyPipewrench (dismissed):

luckyPipewrench (approved):

PR #767, opened 2026-06-14
test(certgen): make read-only-dir tests portable

annawhooo:
## What

Skips two `internal/certgen` tests on Windows. Fast follow-up to #766 (same root cause, separate scope).

`TestWriteCAFiles_ReadOnlyDir` and `TestWriteCAFiles_CertDirCreateFails` call `os.Chmod(roDir, 0o444)` to make a directory read-only so a subsequent `MkdirAll` under it fails. On Windows, POSIX mode bits don't gate directory writes, so `MkdirAll` still succeeds and `writeCAFiles` returns `nil`. `CertDirCreateFails` then panics dereferencing the nil error:

```
panic: ... certgen_test.go:677  (strings.Contains(err.Error(), "create cert directory"))
```

There's no portable way to force a read-only-directory write failure via `chmod` on Windows, so both are skipped there. The error paths stay covered on Unix CI. Same rationale as #711 and #766.

`TestWriteCAFiles_WriteCertFails` (just below these) forces its failure with a portable trick — creating a directory where the cert file should be written — so it is unaffected and keeps running on Windows.

## Verification

- Before: both FAIL on `windows/amd64` (`CertDirCreateFails` panics on nil-error deref). After: both SKIP.
- `go test ./internal/certgen/` full package: `ok`.
- `go vet ./internal/certgen/`: clean.
- `-race` and `golangci-lint`/`gofumpt` run in CI on Linux (no gcc / tools not installed on the Windows dev box). `runtime` was already imported in this file, so no import changes.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform reliability of certificate generation tests by replacing permission-based directory checks with a platform-independent setup to reliably trigger key/certificate directory-creation failure scenarios, ensuring consistent behavior on Windows and Unix-like systems.


luckyPipewrench (changes requested):
Similar to 766, could we keep these two tests running on Windows instead of skipping? For `writeCAFiles`, the failing operation is `MkdirAll(filepath.Dir(path))`, and there is a portable way to force that: create a regular file at the would-be directory path, then pass a child path under it.

For example:
- key-dir failure: create `blocker := filepath.Join(dir, "not-a-directory")`, write a file there, then use `keyPath := filepath.Join(blocker, "subdir", "ca-key.pem")`
- cert-dir failure: same blocker pattern, with `certPath := filepath.Join(blocker, "subdir", testCertFile)`

That should fail on Windows/macOS/Linux without relying on POSIX chmod bits, so we keep the error-path coverage everywhere.

luckyPipewrench (dismissed):

luckyPipewrench (approved):

PR #891, opened 2026-07-04
Harden scanner coverage: SSRF surfaces, encoded secrets, FP precision

luckyPipewrench:
## What changed

Extends scanner coverage across surfaces that were under-scanned, closes several encoding-based evasion gaps, and tightens false-positive precision without weakening detection.

- **A2A**: the agent-card `url` field and message file-part `uri` are now modeled and routed through the URL/SSRF scanner (previously only name, description, and skill text were scanned). The card `url` is included in the agent-card identity hash.
- **Embedded URLs in DLP text**: `http(s)` URLs embedded in scanned text (such as tool arguments) are evaluated through the URL/SSRF scanner across raw, JSON-escaped, and bounded percent-decoded views, so an internal or metadata-host URL is caught. Infrastructure and protective failures no longer over-block otherwise-clean text.
- **HTML-entity decoding in DLP**: text and body DLP now HTML-entity-decode before matching, as a bounded fixpoint that resolves nested encodings, composing with the existing base64/hex decode passes.
- **Response injection**: the credential-exfil pattern now covers markdown link forms (inline, nested-label, angle-bracket, reference-style) alongside the existing image form, with parity maintained across all presets and canonical goldens.
- **MCP tool scanning**: tool names that collide after confusable normalization are flagged as poisoning, catching cross-script look-alike impersonation of a legitimate tool.
- **Entropy / false-positive precision**: the query-value entropy check no longer flags credential-less database connection strings or short human-readable hyphenated values, within a tightly-bounded entropy band; a database URI aimed at an IP-literal or cloud-metadata host now blocks. Opaque high-entropy tokens and credential-bearing connection strings continue to block.

Regression tests accompany each change.

## Summary by CodeRabbit

* **New Features**
  * Added a new credential-exfiltration response rule that detects sensitive language combined with Markdown/URL-style `https://` links (warn-only or blocking depending on preset).
  * Improved embedded URL scanning with stricter allow/deny gating and explicit truncation behavior.
  * Expanded A2A method recognition and enhanced Agent Card URL handling for scanning/identity.
  * Added detection for Unicode “confusable” tool-name collisions.

* **Bug Fixes**
  * Strengthened DLP and response scanning for deeply HTML-entity-encoded content.
  * Refined query handling: allow constrained “safe” DSN forms and readable hyphenated phrases, while blocking unsafe DSN and metadata targets.

* **Tests**
  * Added extensive regression coverage for the new detections, SSRF blocking, HTML decoding, embedded-URL truncation, and edge-case parsing.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
All three findings addressed in the latest commit:

1. **(high, WebSocket)** Reverted the unexpected-continuation tolerance. An unexpected continuation frame is a strict protocol error again on both paths, and the proxy no longer accepts protocol-invalid framing.

2. **(medium, scan API)** Embedded-URL scanning now collects all findings up to a bounded cap instead of returning only the first, and it caps the number of synchronous scans per request.

3. **(medium, scan API)** Deterministic URL denials, including protective and policy denies, now surface in the response, and only true scanner-execution failures are suppressed.

Celest-Blue (approved):

PR #895, opened 2026-07-04
fix(mcp): audit completeness, full-object drift, fail-closed hardening

luckyPipewrench:
Three related hardening changes for the MCP path.

Receipt audit completeness: when a block decision cannot emit its receipt under required-receipts mode, the failure is surfaced with a distinct warning across all block-receipt paths instead of a generic error. The block always fires regardless of receipt emission; this only closes an audit-trail gap.

Drift detection: tool rug-pull drift now hashes the full canonical tool object (annotations, output schema, sibling metadata), matching provenance-attestation scope, so an annotation change is caught even when provenance is not configured. Key order and whitespace do not affect the digest.

Fail-closed hardening: a required receipt with no action id now fails closed; the playground broker sets X-Content-Type-Options nosniff on static assets; and the orphan-VM reaper raises a distinct alert when it repeatedly finds zero managed VMs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved receipt-emission failure handling: blocked outcomes remain correct while “audit gap” logging now appears only when receipts are required.
  * Added fail-closed behavior for missing required receipt details and made receipt/audit logging more consistent.
  * Strengthened provenance-aligned tool hashing to reduce drift false-positives from JSON formatting differences.

* **New Features**
  * Added an alert when the broker repeatedly reports zero managed machines for several reconciliations.

* **Other**
  * Updated static asset responses to include `X-Content-Type-Options: nosniff`.


Celest-Blue (approved):

luckyPipewrench:
@coderabbit review

PR #899, opened 2026-07-04
fix(proxy): scan over-cap size-exempt responses with bounded memory

luckyPipewrench:
## What changed

Over-cap responses from `response_scanning.size_exempt_domains` hosts previously streamed to the client without being scanned. An attacker who controls or injects into a size-exempt upstream could pad past the scan cap and place injection, secret exfiltration, or malware after the scanning window.

Over-cap bodies from size-exempt hosts are now read into memory up to `response_scanning.size_exempt_scan_max_bytes` (default 64 MiB) and run through the existing whole-buffer scanner before any bytes reach the client. Delivery is blocked fail-closed above the ceiling, over a process-wide `response_scanning.size_exempt_scan_max_inflight_bytes` budget (default 256 MiB), or on any read or scan error. A shared helper applies this identically across the forward, CONNECT, and reverse transports.

Adds `response_scanning.unscannable_passthrough` as the only sanctioned unscanned path for opaque-by-construction hosts: a structured allowlist keyed on host, segment-bounded path prefix, and content type, with a required reason and expiry, audit-warned and receipted on every use. Expired, malformed, or non-matching entries fall back to scanning.

## Summary by CodeRabbit

* **New Features**
  * Added size-exempt response-scanning limits (per-response and in-flight) and expanded handling for intentionally unscannable attachment-like responses via `unscannable_passthrough`.
* **Bug Fixes**
  * Improved fail-closed blocking for over-ceiling and in-flight budget exhaustion, with safer fallback when passthrough criteria don’t fully match.
  * Enhanced receipt/block reporting for response-scanning, injection, and strip-failure scenarios.
* **Documentation**
  * Updated response-scanning configuration docs with new knobs, stricter rules, and matching/audit semantics.
* **Tests**
  * Expanded validation and end-to-end forward/reverse proxy regression coverage.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #900, opened 2026-07-04
fix(mcp): enforce denial-of-wallet and chain detection on A2A methods

luckyPipewrench:
## What changed

A2A methods (such as SendMessage and tasks/send) were exempt from denial-of-wallet cost control and tool-chain sequence detection, because those gates keyed only on tools/call tool names. An agent could drive unbounded A2A calls, or run a malicious A2A call sequence that the equivalent tool calls would have blocked.

This introduces a shared enforcement identity (the raw tool name for tools/call, a namespaced `a2a:<method>` identity for A2A methods) and applies denial-of-wallet and tool-chain detection to both on the HTTP and stdio evaluation paths, preserving the existing ordering contract that a denied call leaves no chain trace.

It also adds A2A request-body scanning on stdio for parity with HTTP, and records the cross-agent contamination signal before the short-circuiting gates even when A2A body scanning is disabled.

## Summary by CodeRabbit

* **New Features**
  * Expanded input enforcement to cover A2A actions alongside standard tool calls.
  * Added support for blocking and warning on A2A body metadata URL scans.
  * Improved audit and log attribution so enforcement events are reported with clearer action identity.

* **Bug Fixes**
  * Fixed false negatives in chain and wallet-budget enforcement for A2A traffic.
  * Preserved cross-agent escalation records even when later checks block a request.
  * Prevented blocked A2A requests from being forwarded.

### DIFF
--- a/internal/mcp/a2a_enforcement_parity_test.go
+++ b/internal/mcp/a2a_enforcement_parity_test.go
@@ -201,6 +201,60 @@ func TestForwardScannedInput_A2ADoWBlock(t *testing.T) {
 	}
 }
 
+func TestForwardScannedInput_A2ADoWBlockLeavesNoChainTrace(t *testing.T) {
+	sc := testInputScanner(t)
+	opts := testOpts(sc)
+	opts.InputCfg = &InputScanConfig{Enabled: true, Action: config.ActionWarn, OnParseError: config.ActionBlock}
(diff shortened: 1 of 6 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #909, opened 2026-07-05
fix(mcp): enforce session-binding, taint, and contract gates on A2A methods

luckyPipewrench:
## What changed
A2A (agent-to-agent) JSON-RPC methods are identified by method name and carry no tool name, so they bypassed the tools/call-scoped session-binding, taint action-classification, and live-lock contract gates. This routes A2A methods through all three gates using their namespaced `a2a:<method>` enforcement identity, on both the HTTP and stdio MCP transports, failing closed so a blocked method never reaches the upstream.

Frozen-tool inventory stays tools-only with an explicit documented exclusion and test: A2A methods are not members of a server's frozen tool set, so they are covered by the method-scoped binding, taint, chain, and denial-of-wallet gates instead.

A tool literally named `a2a:<method>` is kept distinct from the A2A method identity so it cannot impersonate it in the newly wired gates.

## Why
Closes a bypass class where A2A traffic escaped enforcement that already applied to tools/call, the same method-based-A2A-skips-tool-name-scoped-gates family as the recently landed baseline, denial-of-wallet, and chain-detection A2A fixes. This extends it to the remaining session-binding, taint, and contract gates.

## Summary by CodeRabbit

* **New Features**
  * Added A2A method identity handling across enforcement, HTTP input scanning, and stdio forwarding.
  * Introduced collision-safe callable identity and callable-args extraction for consistent taint and session-binding decisions.

* **Bug Fixes**
  * Corrected session-binding capture, audit, and block messaging to reflect the enforced target consistently.
  * Ensured taint and session-binding denials are blocked with correct reasons and are not forwarded upstream/subprocess.

* **Tests**
  * Expanded HTTP listener proxy, scanner, and stdio parity tests for A2A “no baseline/unknown tool” and “ask denied” cases.
  * Added live-lock and contract gate coverage validating A2A namespacing and same-name MCP non-permission.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (dismissed):

Celest-Blue (approved):

PR #916, opened 2026-07-05
feat(scanner): operator-configurable endpoint/parameter query-entropy exemption

luckyPipewrench:
## What changed

Adds a neutral, operator-configurable way to relieve query-value entropy false positives on a specific endpoint and parameter, without exempting an entire host.

- New `fetch_proxy.monitoring.query_entropy_param_exclusions` config list. Each entry names an exact scheme (https), DNS host, canonical escaped path, and a single query parameter, and a matching request has only the raw Shannon-entropy heuristic suppressed for that one parameter value.
- Every other scanner still runs on the value and URL: DLP and core DLP, SSRF including the database-URI guard, scheme, path traversal, path and subdomain entropy, adjacent-parameter scanning, and rate and data budgets. A secret in the exempted parameter, or in any adjacent parameter, still blocks.
- Strict validation rejects broad or ambiguous entries (wildcards, ports, userinfo, IP-literal hosts, non-canonical or over-long hosts and paths, http scheme, duplicate tuples). Runtime host matching is IDNA-normalized so a Unicode host matches its canonical punycode config entry, and fails closed to no-match on any malformed host.
- `pipelock doctor` and `pipelock check` emit advisories for missing or expired lifecycle metadata (reason, owner, expires) and for an entry already covered by a host-wide exclusion. Config reload warns when an exemption tuple is added or removed. The effective tuples participate in the canonical policy hash in sorted order, while advisory metadata does not.
- The existing host-wide `query_entropy_exclusions` remains as the broader fallback. Documentation uses neutral placeholder hosts.

## Summary by CodeRabbit

* **New Features**
  * Added endpoint-specific query-entropy exclusions (scheme/host/path/parameter) to allow high-entropy query *values* for exact targets while preserving other protections.
* **Bug Fixes**
  * Improved config canonicalization and stable reload/diagnostic behavior, including warnings for inert, redundant, missing, malformed, and expired exclusions.
* **Documentation**
  * Updated Fetch Proxy monitoring docs and remediation guidance with precise matching rules and knob fallback behavior.
* **Tests**
  * Expanded validation, canonical-hash, reload-warning, scanner matching, and explain/doctor coverage for the new exclusions.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #962, opened 2026-07-10
fix(reverse): label unscanned media passthrough honestly, never clean

luckyPipewrench:
## What changed

Label unscanned media passthrough honestly so no surface reads it as clean. When the reverse proxy streams a declared audio or video response, or a sniffed image or generic binary response, it does so without running the text-injection scanner. It now records that outcome as `media_passthrough_unscanned` (boundary-limited) instead of a scanned-implying reason, so a receipt, coverage certificate, or dashboard view cannot present an unscanned media stream as clean or complete.

The exposure this makes visible: an upstream can serve instruction-bearing text under an audio or video content type that a downstream agent later transcribes and acts on. This change is labeling only; it does not fail-close media, add transcription, or change any scanning verdict.

This also tightens the response-scan audit and config wording so "skipped" is said only where scanning genuinely did not run, distinct from paths that scan and then pin to warn.

This is scoped to the reverse proxy. The same passthrough over-claim exists in other transports (forward, TLS intercept, WebSocket) and belongs to the separate transport-parity media-trust work; those are flagged, not changed here.

## Verification

Proven by neutralization: reverting the outcome reason to a scanned-implying value fails the media passthrough tests, which cover audio, video, image, and generic-sniffed media plus the unscannable-passthrough capture outcome. Build and lint clean; race tests green across proxy, audit, and config.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response-scan exemption audit/log semantics to clearly separate warn-only handling from full-trust streaming bypass behavior.
  * Refined the configuration warning message shown when exemptions are set while response scanning is disabled.
  * Updated media/binary passthrough telemetry to consistently label unscanned media passthrough, and marked unscannable passthrough capture outcomes as skipped.
* **Tests**
  * Updated response-scan exemption assertions for the new “pinned to warn” semantics.
  * Added/updated receipt verification tests for unscanned media passthrough labeling and skipped capture outcomes for unscannable passthrough.

### DIFF
--- a/internal/audit/logger.go
+++ b/internal/audit/logger.go
@@ -735,10 +735,10 @@ func (l *Logger) LogAnomaly(ctx LogContext, scanner, reason string, score float6
 	}
 }
 
-// LogResponseScanExempt logs when response injection scanning is skipped because
-// the destination host matched response_scanning.exempt_domains. This is a dedicated
-// event type (not "anomaly") so SIEM consumers can filter exemption events separately
-// from actual security anomalies.
+// LogResponseScanExempt logs response_scanning.exempt_domains handling for the
+// generic path, where scanning still runs but findings are pinned to warn. This
+// is a dedicated event type (not "anomaly") so SIEM consumers can filter
+// exemption events separately from actual security anomalies.
 func (l *Logger) LogResponseScanExempt(ctx LogContext, hostname string) {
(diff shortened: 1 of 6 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #974, opened 2026-07-11
feat(dashboard): OIDC bearer-token authentication

luckyPipewrench:
Adds OIDC bearer-token authentication as an additive authenticator for the enterprise dashboard, alongside the existing operator token. Operators choose either or both.

## What changed

- New flags on `dashboard serve`: `--oidc-issuer` (issuer URL, discovered to a JWKS endpoint), `--oidc-audience`/`--oidc-client-id` (expected audience), `--oidc-role-claim` (the claim carrying roles or groups), and `--oidc-role-map` (maps a claim value to a role and a bounded dashboard permission set, including raw access).
- Verifies the presented bearer token's RS256 signature against the issuer JWKS on every request, with cached JWKS refresh, and enforces issuer, audience, expiry, not-before, and optional authorized-party claims with a small clock-skew leeway.
- The algorithm is pinned to RS256 and only parsed RSA keys are used, so unsigned and algorithm-substitution tokens are rejected; token claims are read only after the signature verifies.
- Discovery and JWKS are treated as fixed trust anchors and fetched without following redirects.
- Maps the verified role claim to a bounded permission set and fails closed on any verification failure, excessive clock skew, or unmapped role; an empty role map is a startup error.
- Additive and non-weakening: token authentication is unchanged, and the existing token/raw split and non-loopback-requires-TLS guard are preserved.
- Stdlib-only implementation; no new dependencies. Enterprise build tag; the non-enterprise build is unaffected.

## Notes

This and the mTLS authenticator PR both extend the same authorizer-construction block in `enterprise/cli/dashboard.go`; whichever merges second will need a small rebase to compose all three identity sources (token, mTLS, OIDC).

## Summary by CodeRabbit

* **New Features**
  * Added OIDC authentication for the dashboard, mapping JWT roles to dashboard permissions.
  * Added enterprise mTLS authorization using client-certificate fingerprints to grant permissions and raw access.
* **Security / Improvements**
  * Tightened TLS and client-certificate flag validation and fail-closed behavior when certificate auth is enabled.
  * Improved authentication failure handling and enhanced audit/access logging with sanitized auth context.
* **Documentation**
  * Updated dashboard CLI docs to cover OIDC and mTLS authentication, redaction behavior, and required flags.
* **Tests**
  * Added end-to-end OIDC and mTLS test coverage for validation, caching, routing, and permission gating.

### DIFF
--- a/docs/cli/dashboard.md
+++ b/docs/cli/dashboard.md
@@ -68,7 +68,7 @@ openssl rand -hex 32 > /etc/pipelock/dashboard.token
 |---|---|---|
 | `--receipt-dir` | (required) | Flight-recorder evidence directory holding action receipts (the runtime's `flight_recorder.dir`). |
 | `--config` | none | Optional Pipelock config file for the read-only Exemptions inventory. When omitted, `/exemptions` renders an explicit "no config loaded" state and the Evidence view still works. |
-| `--auth-token-file` | (required) | File containing the operator token required on every request. Grants the redacted metadata view. |
+| `--auth-token-file` | (required) | File containing the operator token for token-authenticated requests. Grants the redacted metadata view. |
 | `--raw-token-file` | none | Optional second, higher-privilege token that unlocks raw destinations and signed payloads. Must differ from `--auth-token-file`. |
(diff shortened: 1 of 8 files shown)


luckyPipewrench:
Superseded by #973 after collapsing the dashboard auth stack into one PR.

PR #1020, opened 2026-07-17
build: add CI-equivalent test shards to the Makefile

luckyPipewrench:
## What changed

The only full-suite Make target ran the whole race suite as one monolithic all-packages invocation. This adds per-shard targets that mirror the CI package split (`scripts/ci_test_packages.py`: proxy, scanner, mcp, rest): `test-shard-<name>`, `test-shard-enterprise-<name>`, `test-sharded`, and `test-sharded-enterprise`, each carrying the same 15m timeout CI uses.

A developer can now reproduce a single CI shard locally, or run the full suite in scoped, labelled chunks instead of one long-pole invocation. The single-shard pattern targets take a `FORCE` prerequisite so they always run. The existing full-suite target is unchanged and remains the canonical full local run.

## Summary by CodeRabbit

* **Tests**
  * Added CI-equivalent sharded test targets for standard and enterprise configurations.
  * Added support for running individual test shards or all shards sequentially.
  * Sharded tests run with race detection and stop when a failure occurs.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1032, opened 2026-07-18
fix(config): make internal/config tests build on Windows

Celest-Blue:
`internal/config` tests could not compile on Windows: `config_test.go` called `syscall.Mkfifo`, which does not exist there. Because the failure was at compile time, the package's existing `t.Skipf` fallback could never run, and every test in the package was unreachable on Windows — including the ones `make docs-check` and `make stats` depend on.

The Windows lane did not catch this: `go build ./...` never compiles `_test.go` files, and the targeted `go test` steps cover five other packages, none of which is `internal/config`.

## Changes

- **Split the FIFO case by build tag.** `license_nonregular_unix_test.go` keeps the FIFO, which is what `resolveLicenseKey` specifically guards against since a FIFO can block the reader. `license_nonregular_windows_test.go` uses a directory instead — Windows has no `mkfifo`, but a directory is equally non-regular, so it exercises the same `Mode().IsRegular()` rejection rather than dropping the coverage. `license_nonregular_test.go` holds the shared assertion.
- **Wire it into the Windows lane** so this cannot regress silently. Compiling the package is the actual guard; `go build ./...` cannot provide it. The stale lane comment claiming `internal/config` uses unix-only syscalls is updated.
- **Force LF for `*.md` and `*.go`.** Git for Windows defaults `core.autocrlf=true`, and a CRLF checkout breaks two gates with no real content drift: `TestDocsDeclareLiveStatsAndDefaults` matches multi-line literals containing `\n` (`docs_accuracy_test.go:26`), and gofumpt treats a CRLF file as unformatted, so `make lint` reports every Go file in the repo as needing formatting. This follows the existing `**/testdata/**` rule, present for the same reason.

No committed `.go` or `.md` blob contains CRLF today, so these rules normalize nothing and produce no renormalization diff. On Linux they are a no-op.

## Scope

This makes the package compile and fixes the doc and lint gates. **27 tests in `internal/config` still fail on Windows** — they assert Unix permission-model behavior (world-readable CA key, group-writable `license_file`, owner-executable `0o700`, private temp base). Those need either explicit Windows skips or ACL equivalents, which is a separate decision and is left untouched here.

## Verification

- `make docs-check` passes on Windows
- `golangci-lint run ./internal/config/...` — 0 issues
- `gofumpt -l .` clean
- `go vet ./...` clean repo-wide
- the new CI step runs and passes: `go test -count=1 -run TestWindows ./internal/config/...`

## Open question

The `.gitattributes` fix is surgical (`*.md`, `*.go`). A single `* text=auto eol=lf` would prevent the next instance of this class. Happy to switch if maintainers prefer.

## Summary by CodeRabbit

* **Tests**
  * Added Unix and Windows coverage to ensure license key paths that aren’t regular files (e.g., FIFOs or directories) are rejected with the expected error.
  * Updated the Windows CI lane to compile and run the Windows-specific configuration test subset.
* **Chores**
  * Standardized line endings for Markdown and Go files to prevent cross-platform formatting and byte/content matching issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,3 +7,14 @@
 # Windows checkouts get CRLF and tests fail byte-for-byte against
 # the LF-committed goldens without any real content drift.
 **/testdata/** text eol=lf
+
+# Markdown is content-asserted by TestDocsDeclareLiveStatsAndDefaults, which
+# matches multi-line literals containing "\n". A CRLF working-tree checkout
+# breaks those matches on Windows for the same reason as the goldens above,
+# with no real content drift.
+*.md text eol=lf
+
+# gofumpt and the gofmt-family linters treat a CRLF file as unformatted, so a
+# Windows checkout with core.autocrlf=true (the Git for Windows default) makes
+# `make lint` report every Go file as needing formatting.
+*.go text eol=lf
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -233,10 +233,14 @@ jobs:
   # (#696) regress on real Windows undetected. This is a TARGETED lane, not full
   # parity: the full suite asserts Unix file modes that Go can't report on
   # Windows (it derives FileMode from the read-only attribute, never the ACL),
-  # and internal/config test files use unix-only syscalls. So we run the
-  # OS-split predicate plus the load/capture paths that #695/#696 touch, scoped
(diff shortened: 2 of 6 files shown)


luckyPipewrench (approved):
Great to see windows get some love!

PR #1055, opened 2026-07-21
feat(mcp): add data-class label taxonomy and config foundation

luckyPipewrench:
## What changed

Adds the foundation for DLP-derived data-class labels on MCP receipts: a new `internal/datalabel` package with a closed `DataClass` taxonomy (benign, credential, secret, hash, network, identity, pii) and a pure `DataClassFor(redact.Class)` mapping that fails closed to `secret` for any unknown class. Adds a disabled-by-default `mcp_data_class_labels` config section with full load, validation, and hot-reload handling across all presets.

## Scope

This is the taxonomy and config surface only. It has no runtime effect yet: nothing derives or attaches labels, and the receipt fields stay empty. The derivation path that reads scanner findings and populates the receipt `data_classes_in`/`data_classes_out` fields lands in a follow-up. The config knob is inert by design until then, and it stays out of the canonical policy hash so enabling it does not change signed receipt bytes.

## Why split it out

Landing the taxonomy, the fail-closed mapping, and the config lifecycle first keeps the follow-up focused on the derivation wiring in the scan path, which is the higher-risk change. Labels are always synthesized internally from scanner findings and are never parsed from an agent payload.

## Summary by CodeRabbit

* **New Features**
  * Added an MCP “data-class labels” configuration block across presets, disabled by default.
  * Introduced a fail-closed fallback label for unknown classifications (`secret`).
  * Added configuration schema and validation (defaults, allowed values, and rejection of `enabled: true` as not supported yet).

* **Bug Fixes**
  * Ensured the disabled setting does not affect stable policy fingerprints.
  * Added a reload warning when data-class labeling is turned off between configurations.

### DIFF
--- a/configs/audit.yaml
+++ b/configs/audit.yaml
@@ -388,6 +388,10 @@ mcp_tool_scanning:
   action: warn
   detect_drift: true
 
+mcp_data_class_labels:
+  enabled: false
+  unknown_class: secret  # foundation only; no runtime effect until derivation wiring lands
+
 mcp_tool_policy:
   enabled: true
   action: warn
--- a/configs/balanced.yaml
+++ b/configs/balanced.yaml
@@ -424,6 +424,10 @@ mcp_tool_scanning:
   action: warn
   detect_drift: true
 
+mcp_data_class_labels:
+  enabled: false
+  unknown_class: secret  # foundation only; no runtime effect until derivation wiring lands
+
 mcp_tool_policy:
   enabled: true
   action: warn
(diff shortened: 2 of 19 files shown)


luckyPipewrench:
Dispositions on this review:

1. `mcp_data_class_labels.enabled` now fails closed at load until label derivation is wired, and the inert field was removed from the reload-downgrade warning, so the knob can no longer advertise a protection it does not provide.

2. The config-consumption checker matches bare top-level field names, so a path-qualified allowlist entry would not match today. `UnknownClass` has no other struct in the tree, so there is no current masking; I'll qualify it if the checker gains path support.

DeathCamel58 (approved):

PR #1061, opened 2026-07-21
feat(emit): OTLP sink delivery-failure accounting and health stats

luckyPipewrench:
The OTLP audit-event sink swallowed every delivery failure to stderr: a worker-goroutine panic killed the goroutine silently, a drain timeout abandoned queued events without counting them, and marshal, compression, request, send, and non-2xx errors never surfaced in any metric, so a dropped OTLP event was invisible to operators.

This gives the OTLP sink the same delivery accounting the syslog sink already exposes: a Stats snapshot (delivered, failed, dropped, abandoned, degraded, last error, queue depth), a degraded flag that clears on the next successful export, per-event panic recovery so one bad event no longer takes down the worker, and abandoned-event counting on drain timeout and close-during-retry. Emit returns a degraded advisory after a prior async failure and counts queue-full drops. Structured stderr diagnostics carry no event content.

## Summary by CodeRabbit

* **New Features**
  * Added exported OTLP delivery health statistics (delivered/failed/dropped/abandoned), queue info, degraded state, and last error snapshot.
  * OTLP exports now surface degraded and queue-full conditions (including an advisory degraded error on successful enqueue).
  * Improved export diagnostics with structured JSON error reporting.
* **Bug Fixes**
  * More robust failure, retry, panic, and shutdown/drain handling to ensure accurate accounting and continued processing where appropriate.
* **Tests**
  * Added comprehensive accounting/diagnostics tests, including edge cases, retry outcomes, queue precedence, drain timeout abandonment, and stderr behavior.


luckyPipewrench:
Addressed in `b46ae1a1`.

**#2 (synchronous stderr on the Emit hot path):** `recordDropped` is now counter-only on the hot path, so a queue-full drop is observable via `Stats()` (dropped, degraded, last_error) instead of a synchronous stderr write that could block the request path when the sink is under backpressure and stderr is stalled.

**#3 (stale `LastError`):** a successful delivery now clears `LastError` when it clears `Degraded`, so a non-degraded snapshot no longer reports an old error.

**#1 (`Emit` returns `ErrOTLPDegraded` on successful enqueue):** kept intentionally. It mirrors the existing syslog sink contract, and the sole consumer of a sink's `Emit` (the emitter fan-out, via the transparent filter wrapper) logs the error and continues, with no retry and no fail-closed, so the advisory return is safe. The `Emit` doc now states this explicitly: `ErrOTLPDegraded` means the event was accepted, not rejected. Bringing the syslog and webhook sinks to the same hot-path and `LastError` behavior is tracked as a follow-up so the sink family stays consistent.

DeathCamel58 (approved):

PR #1072, opened 2026-07-22
fix(scanner): carve verified inline images out of the query-entropy gate

luckyPipewrench:
## What changed

A base64-encoded image passed as a URL query value (data:image/png;base64,...) is high-entropy by construction and tripped the query-parameter entropy block, a false positive for agents that legitimately pass inline images. This skips the entropy check for a query value that is exactly a verified image, reusing the verified-image recognition the response and text-DLP scanners already apply.

The carve-out is deliberately narrow so it cannot become an exfiltration channel: it accepts only a strict true-color PNG whose bytes are fully accounted for (an exact data:image/png;base64 header with no MIME parameters, a single IDAT that is a completely-consumed zlib stream, and no ancillary, private, or palette chunks) and decodes it to confirm. JPEG and indexed PNG fail closed because their decoders ignore metadata segments that could smuggle data.

A non-image payload, a fake image envelope, a secret placed in a header parameter or before, after, or around the image, and secrets split across separators all remain blocked and entropy-scanned. The only residual is a secret steganographically embedded in real pixel data, the same accepted limitation the existing image-excision scanners already carry.

## Notes

Every carve-out and block is verified by reproduction, including the case where the recognizer is stubbed and a fake payload must still be blocked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query entropy scanning for embedded PNG/JPEG base64 `data:image/...` values, preventing verified image payloads from being treated as high-entropy.
  * Added stricter verification for eligible `data:image/png;base64` and `data:image/jpeg;base64` data URLs, requiring complete decoding and strict PNG acceptance (with dimension limits).
  * Protected verified image payloads from being split on query delimiters and blocked entropy checks on protected placeholders.
* **Tests**
  * Expanded regression coverage for allowed/rejected image boundaries, high-entropy “fake” image-shaped payloads, and placeholder injection via URL input.

### DIFF
--- a/internal/scanner/opaque_image_dataurl.go
+++ b/internal/scanner/opaque_image_dataurl.go
@@ -120,6 +120,12 @@ func stripVerifiedImageDataURLs(text string, retainDecoded bool) (excised, decod
 	)
 }
 
+// stripVerifiedImageDataURLsMatching scans text for data:image base64 URLs and
+// removes (or replaces) only those whose header satisfies headerAllowed and
+// whose decoded bytes satisfy decodedAllowed. It returns the stripped text and,
+// when retainDecoded is set, the concatenation of the decoded image bytes. The
+// header/decoded predicates let callers apply a stricter policy (the
+// query-entropy carve-out) than the response/text-DLP excision.
 func stripVerifiedImageDataURLsMatching(
 	text string,
 	retainDecoded bool,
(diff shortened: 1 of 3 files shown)


DeathCamel58 (approved):

luckyPipewrench:
Shelving after an adversarial design review: this query-entropy image carve-out weakens an outbound anti-exfil control by default to address a low-frequency false positive that is already remediable with the narrower, operator-controlled query_entropy_param_exclusions knob. Recorded as decided do-not-build; may revisit only as a typed, default-off, per-endpoint exclusion if real false-positive prevalence is measured.

PR #1106, opened 2026-07-26
feat(scanner): detect opaque high-entropy content on body, WebSocket, and A2A egress

luckyPipewrench:
## What changed

Pattern-based DLP only catches exfiltration that matches a known credential signature. A stolen canary-file value, an opaque session token, or a hex-encoded blob carries no signature, so a pattern scanner never sees it leave. URL transports already had entropy detection for this; request bodies, WebSocket client-to-server frames, and A2A message bodies did not.

This adds a shared opaque-content-entropy check to those egress paths. A high-entropy value leaving to a destination that is not trusted is flagged even when it matches no DLP pattern. The detector lives in `internal/contententropy` and is shared by the proxy body-scan path and the A2A scanner, so there is one implementation rather than two.

## Design

Detection is destination-gated, and the destination is the parsed upstream authority, never a request-supplied Host header. Hosts in `trusted_domains` or `content_entropy_exclusions` are skipped. Entropy suppression for a trusted destination never suppresses a DLP match in the same body, so trusting a host for opaque content does not weaken credential detection.

New config lives under `request_body_scanning` and `websocket_proxy`: `content_entropy_enabled`, `content_entropy_action` (warn or block), `content_entropy_threshold` (Shannon bits per character), `content_entropy_min_length`, and `content_entropy_exclusions`. The shipped default is warn in the general presets so the detector observes and audits before it enforces; the strict and hostile-model presets block.

## False-positive posture

Body and frame content is high-entropy by nature (base64 uploads, encrypted payloads, content-addressed hashes), so a naive always-block detector would reject legitimate traffic. The warn default plus destination-gating plus a minimum-length floor keep it quiet out of the box, and operators tune a specific endpoint with the narrow per-host exclusion before switching it to block. The configuration guide and the false-positive tuning guide document the knobs and the narrowest-first workflow.

## Known limits, documented

A content-addressed hash and a hex-encoded secret are the same shape, so exact-length hex is treated as opaque content rather than declared safe by length (declaring digest lengths safe was a bypass: an attacker sizes the payload to a digest length). In block mode this can flag a legitimate hash upload, which is why the general default is warn and block-mode presets are tuned by destination. WebSocket detection is per complete message; a blob split across separate messages under the minimum length is left to the cross-request data budget. MCP-transport A2A gateways are not covered yet because those paths lack a parsed upstream authority to gate destination trust on; HTTP and CONNECT-proxied A2A message bodies are covered.

## Tier

Free. This is single-agent detection with no paid aspect.

## Summary by CodeRabbit

* **New Features**
  * Added opaque/high-entropy content detection for request bodies, A2A payloads, and WebSocket per-message bodies.
  * Introduced configurable warn/block enforcement (entropy threshold, minimum length gate) with per-destination entropy exclusion lists; supported both HTTP and WebSocket flows.
  * Added `body_entropy` audit events, block-reason code/layer, and remediation guidance.
  * Exposed new body-entropy audit telemetry and metrics.
* **Documentation**
  * Documented new configuration fields, false-positive tuning, and `body_entropy` block-reason header vocabulary.
* **Bug Fixes**
  * Improved interaction rules so entropy findings aren’t masked/downgraded by other findings or budget-exceeded outcomes.

### DIFF
--- a/configs/audit.yaml
+++ b/configs/audit.yaml
@@ -40,6 +40,11 @@ forward_proxy:
 request_body_scanning:
   enabled: true
   action: warn
+  content_entropy_enabled: true
+  content_entropy_action: warn
+  content_entropy_threshold: 4.5
+  content_entropy_min_length: 32
+  content_entropy_exclusions: []
   max_body_bytes: 5242880
   scan_headers: true
   header_mode: sensitive
@@ -62,6 +67,7 @@ websocket_proxy:
   max_connection_seconds: 3600
   idle_timeout_seconds: 300
   origin_policy: rewrite
+  content_entropy_exclusions: []
 
 dlp:
   scan_env: true
--- a/configs/balanced.yaml
+++ b/configs/balanced.yaml
@@ -53,6 +53,11 @@ forward_proxy:
 request_body_scanning:
   enabled: true
   action: warn
+  content_entropy_enabled: true
+  content_entropy_action: warn
+  content_entropy_threshold: 4.5
+  content_entropy_min_length: 32
+  content_entropy_exclusions: []
   max_body_bytes: 5242880
   scan_headers: true
   header_mode: sensitive
@@ -89,6 +94,7 @@ websocket_proxy:
   max_connection_seconds: 3600
   idle_timeout_seconds: 300
   origin_policy: rewrite
+  content_entropy_exclusions: []
 
 dlp:
   scan_env: true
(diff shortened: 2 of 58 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (dismissed):

Celest-Blue (approved):

PR #1148, opened 2026-08-02
feat(proxy): bind the SSRF dial rebind snapshot to exact host and port

luckyPipewrench:
## What changed

The forward proxy records a DNS-rebind scan snapshot (the external IPs a host resolved to at scan time) and, at dial time, flags a dialed IP that was not in that set as a rebind. That snapshot was keyed by host only. This binds it to host AND port: the snapshot stores the scanned port, the SSRF-safe dialer records the dial's port in the request context, and the rebind check requires the scan-time snapshot's port to match before it applies.

A new `effectiveURLPort` derives the scheme-default port, and the port is threaded through all six snapshot recording sites: CONNECT, absolute-URI forward, fetch, redirect, intercept, and websocket.

## Why

A snapshot recorded while scanning a host on the default port could previously influence the rebind verdict of a dial to the same host on a different port. This corrects that cross-port coupling and is the exact-port groundwork that upcoming per-service destination grants depend on. The snapshot is label-only today (it refines a block's reason to `ssrf_dns_rebind` and does not gate allow or deny), so this changes no allow or deny verdict.

## Scope

The scanned and contract-evaluated URL is left unchanged so the scanner and contract-gate verdicts stay identical. Whether the CONNECT contract gate should also see the real destination port is a separate design question and is not decided here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SSRF protection by binding DNS-rebinding checks to both the destination host and port.
  * Prevented scan results for one port from being incorrectly reused when connecting to another.
  * Preserved exact destination ports across proxy, redirect, intercepted URL, and WebSocket requests.
  * Maintained support for standard HTTP/HTTPS ports and port-less destinations.
  * Improved consistency of SSRF validation across supported connection types.

### DIFF
--- a/internal/proxy/forward.go
+++ b/internal/proxy/forward.go
@@ -168,6 +168,12 @@ func (p *Proxy) handleConnect(w http.ResponseWriter, r *http.Request) {
 	if strings.Contains(host, ":") { // IPv6 literal needs brackets in URL
 		syntheticHost = "[" + host + "]"
 	}
+	// The scanned/contract URL keeps its historical port-less form so the
+	// contract gate and scanner verdict are unchanged; the exact CONNECT port
+	// is bound to the dial snapshot below (from targetPort) and carried in the
+	// receipt target. Whether the CONNECT contract gate should also see the
+	// real port is a separate design question, tracked as a guard adjacent
+	// finding, not decided here.
 	syntheticURL := "https://" + syntheticHost + "/"
 	connectReceiptTarget := "https://" + net.JoinHostPort(host, targetPort) + "/"
 	targetCtx := newConnectAuditContext(p.logger, target, clientIP, requestID, agent)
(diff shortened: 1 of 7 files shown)


luckyPipewrench:
Context for reviewers: this is the first of the Guard Phase-0 slices (an identity-preserving developer egress guard), so it is intentionally narrow.

The DNS-rebind scan snapshot is label-only today: it refines a block's reason to `ssrf_dns_rebind` and does not gate allow or deny, so binding it to host and port changes no verdict. The value is that the port now travels scan to snapshot to dial, which the upcoming per-service destination grants depend on, and it removes a cross-port coupling where a default-port scan sitting in the request context could influence a different-port dial's rebind verdict.

Deliberately out of scope: the CONNECT scan and contract-evaluated URL is left port-less so the scanner and contract-gate verdicts stay identical. Feeding the real port into the contract gate is a separate design question, since it would subject non-default-port CONNECT targets to non-default-port contract rules, and it is not decided here.

Sibling: #1149 is the other Phase-0 slice up so far, the fail-closed sandbox network lifecycle this feature builds on.

Celest-Blue (approved):

PR #1168, opened 2026-08-04
feat(receipt): specify evidence provenance proofs (spec, typed recipes, hostile corpus)

luckyPipewrench:
## What changed

This adds a normative specification, a typed transform-recipe language, an experimental unregistered proof schema, and a hostile conformance corpus for evidence provenance. It registers nothing, changes no production path, and advertises no capability.

Stacked on #1167 and retargets to main once that merges.

## Why a new shape rather than repairing the existing one

Three defects in the current `proxy_decision_with_spans` shape cannot be fixed in place.

The scanner emits byte offsets while the receipt documents rune offsets, so any wiring between them silently mis-indexes on non-ASCII content with no verifier-visible error. ASCII fixtures conceal it entirely.

The producer, the schema, and the reproducer speak three different view vocabularies. The scanner emits labels the schema does not accept, the verifier can reproduce only eight transforms, and every `dlp_normalized:*` suffix collapses to a single transform, losing path, query, and component semantics.

Per-match commitments leave everything around the match in the view unconstrained.

## The shape

Offsets are UTF-8 byte offsets into a named transformed view. There is no rune conversion. Instead the verifier rejects what a conversion would have silently corrupted: invalid UTF-8, an interval starting or ending inside a code point, negative or overflowing arithmetic, out-of-bounds intervals, and intervals that are unsorted, duplicated, or overlapping.

View labels are replaced by a typed recipe: an ordered sequence of typed operations with explicit parameters, so component selection, repeated query keys, and occurrence are unambiguous. Normalization data, decoding alphabets, padding behavior, maximum decode passes, invalid-input handling, and size limits are pinned by a transform-profile digest. One vocabulary, one implementation contract.

Commitments cover the complete transformed view as well as each match, and every preimage is length-framed and domain-separated, so reordering sources or matches, changing a recipe, or moving a match changes the commitment.

## The commitment stays keyed on purpose

The commitment key is never serialized into the receipt. Publishing an unkeyed digest of undisclosed content would create an offline oracle: an attacker holding a published receipt could confirm a guessed low-entropy secret. That property is preserved deliberately, and the specification says so rather than leaving it as an implementation accident.

## What this does not do

Nothing here is dispatchable. The schema is not in the payload registry, no payload kind changes maturity, and no production code emits any of it. Wiring a producer is separate work that depends on this specification existing first.

## Producer digests are reported honestly

A binary digest or ruleset digest inside a receipt is self-reported by the process that signed it. The specification requires a verifier to report these as attested but unchecked until it independently recomputes them against the artifacts, because a signature proves who asserted a digest and not which binary actually ran.

## Summary by CodeRabbit

* **New Features**
  * Added experimental evidence provenance support for tracking sources, transformed views, and matched intervals.
  * Added deterministic normalization recipes for URL extraction, decoding, text normalization, and character transformations.
  * Added cryptographic commitments to help verify transformed views and matches.

* **Documentation**
  * Added a specification covering provenance terminology, validation, verification states, and conformance requirements.

* **Tests**
  * Added comprehensive validation, transformation, commitment, and conformance coverage.

### DIFF
--- a/docs/specs/evidence-provenance-proof-v1.md
+++ b/docs/specs/evidence-provenance-proof-v1.md
@@ -1,0 +1,47 @@
+# Evidence provenance proof v1 (experimental, fixture-only)
+
+Status: experimental and fixture-only. This document specifies no registered receipt payload, production emitter, verifier command, or capability claim.
+
+## Scope and terms
+
+A source is named input bytes. A view is UTF-8 bytes produced from exactly one source by an ordered typed recipe. A match is a half-open byte interval `[byte_start, byte_end)` in that view. All offsets are unsigned UTF-8 **byte** offsets. Rune/character conversion is forbidden.
+
+Implementations MUST reject invalid UTF-8 source or view bytes; starts or ends in UTF-8 continuation bytes; zero-length, out-of-bounds, or arithmetic-overflowing intervals; unsorted, duplicate, or overlapping intervals. Adjacent intervals are valid. Because the wire shape carries `byte_start` and `byte_end`, a verifier rejects `byte_end <= byte_start` and never derives an end from an unrepresented length field.
+
+## Typed recipe language
+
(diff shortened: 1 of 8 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (dismissed):

Celest-Blue (approved):

PR #1170, opened 2026-08-04
ci: retry externally terminated test shards and narrow build-failure detection

luckyPipewrench:
## What changed

A test shard was killed mid run and the pull request stayed red until a human re-ran it by hand. The job emitted one passing package line, then `Terminated`, then exited 143 after about eight and a half minutes against a fifteen minute test timeout, with no failing test anywhere in its output. Three sibling shards of the same package passed, and the package passed locally under the race detector on the identical commit, so nothing was wrong with the code.

The retry wrapper could not help, by design: it retries only a verified test timeout, so that a genuine failure is never retried away. An externally terminated shard matched no condition, so it was never retried, and its coverage was never uploaded, which can also fail the coverage check on data that has nothing to do with the change under review.

## The fix, in two layers

The primary change reduces the pressure that caused the kill: the two race instrumented packages in that shard now run sequentially rather than concurrently, while keeping the same in-package parallelism.

The backstop retries an exit 143 exactly once, and only when a package started but never reported a terminal result, and the captured output contains no failing test, no failing package, no data race, no build or setup failure, and no panic other than a timeout. Every form of visible failure blocks the retry, including a truncated failure event. Exit 143 on its own is not sufficient, because a command can return it after completing normally or alongside a real failure.

What this can and cannot promise is worth stating plainly. No finite rule can prove a failure never existed in a process that was killed before it could report. The guarantee here is narrower and checkable: no failure visible in the captured output can be swallowed.

Retries announce themselves before and after the rerun, so a resource problem that keeps recurring stays visible instead of being quietly absorbed.

## A separate defect found while in here

The build failure detector matched the bare word `cannot` anywhere in the combined output. That word is common inside ordinary Go error strings and assertion messages, so a plain test failure whose message contained it was misclassified as a build or setup problem. Detection is now limited to compiler and package level output, with a test proving genuine build failures are still caught.

## Coverage

A coverage bearing retry deletes the partial profile, reruns the whole shard, and requires a fresh non empty profile before uploading. If the retry also dies there is no trustworthy coverage to upload. Preventing the coverage service from scoring an incomplete aggregate needs configuration on its side and is not addressed here.

## Verification

The retry decision is tested against each case directly: a real test failure must not retry, a verified timeout must retry, a signal kill with no failure must retry, and a signal kill alongside a failure must not. Neutralising the new predicate flips the positive case back to a red exit, which is how it was confirmed to be doing the work.

## Summary by CodeRabbit

* **Chores**
  * Improved continuous integration test execution by adjusting parallelism for proxy-related test shards.
  * Enhanced test retries with safer process cleanup, clearer failure classification, and better handling of interrupted test runs.
  * Preserved coverage results from initial test attempts for more reliable diagnostics.

* **Tests**
  * Expanded automated coverage for retry behavior, timeout validation, process cleanup, race detection, and coverage preservation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -110,11 +110,18 @@ jobs:
           set -o pipefail
           set +e
           packages="$(python3 scripts/ci_test_packages.py --shard "$TEST_SHARD")" || exit $?
+          package_parallelism=2
+          if [ "$TEST_SHARD" = "proxy" ]; then
+            # The proxy shard has two race-instrumented packages. Running them
+            # sequentially lowers peak memory without serializing t.Parallel
+            # tests inside either package.
+            package_parallelism=1
+          fi
           # Public runners are 4 vCPU, and -race multiplies scheduler pressure.
           # Keep package and intra-package test fan-out below CPU count so
           # deadline/timer tests are not starved by the shard matrix.
           bash scripts/ci-test-with-retry.sh --packages "$packages" -- \
-            go test -race -p=2 -parallel=2 -timeout=15m -count=1 -json \
+            go test -race -p="$package_parallelism" -parallel=2 -timeout=15m -count=1 -json \
             | python3 scripts/stream_go_test_json.py --label "$TEST_LABEL" --json-out /tmp/go-test-oss.json
           test_status=$?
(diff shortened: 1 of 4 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1171, opened 2026-08-04
feat(guard): explain path decisions and fix a component floor bypass

luckyPipewrench:
## What changed

The guard floor refuses credential paths, but an operator had no way to ask why a path was refused, or which rule did it, or whether the refusal was something they could narrow. Two read only commands answer that.

`pipelock guard explain <path>` reports, for one absolute path, whether it would be allowed for read and for write, which rule decides, and what that rule matched. `pipelock guard show <profile>` prints the manifests a profile selects, every path with its declared type, and the floor refusals that apply to them.

## The distinction the output is built around

A refusal from the compiled floor cannot be configured away. A path that is simply not granted can be granted. Those are different problems with different fixes, and an operator who cannot tell them apart will either give up or go looking for a knob that does not exist, so the output separates them explicitly and says so in as many words.

Both commands also state that guard is not enforced in this build. That matters more than it looks: a clean explain output must never leave someone believing a workload is constrained when nothing constrains it yet.

## One implementation of the floor

The validators and the commands consume the same matcher. A second copy of the matching logic would drift from the rules that actually refuse a path, and at that point the command is confidently describing rules the system no longer applies, which is worse than having no command. A test asserts the explained rule is the rule the validator enforced, so the two cannot diverge silently.

## A fail open found by using the command

Driving the finished command against a real symlink surfaced a hole in the floor itself. Forbidden path components were matched only after symlink resolution, so a path declared as a key or credential directory that pointed somewhere innocuous resolved to a name carrying no forbidden component, and was allowed for read and for write alike. Naming such a directory in a manifest is the thing being refused; where the link happens to point today does not make the declaration safe, and it can be repointed afterwards.

The component check now runs against the declared spelling as well as the resolved target, matching what the credential shape check already did, so the refusal also names what the operator actually wrote. The regression test builds the symlink and fails if the declared name stops being checked. An ordinary directory beside the symlinks stays grantable, so the test cannot pass for a blanket reason unrelated to the component.

## Scope

Read only throughout. Neither command writes config, state, or the filesystem, and neither makes a trust decision from anything Pipelock generated. Output goes through the command writer and a stable JSON shape is available alongside the human readable default.

## Summary by CodeRabbit

* **New Features**
  * Added the read-only `pipelock guard explain` command to inspect path decisions.
  * Added `pipelock guard show` to review guard profiles and access declarations.
  * Added human-readable and stable JSON output with matched rules, refusal reasons, and compiled filesystem restrictions.
  * Added validation for paths, profiles, grants, and guard declarations.
  * Clearly indicates that inspection does not enforce guard policies.

* **Documentation**
  * Added comprehensive CLI documentation covering commands, options, output, validation, and exit statuses.

### DIFF
--- a/docs/cli/README.md
+++ b/docs/cli/README.md
@@ -5,6 +5,7 @@ the full command tree and `pipelock <command> --help` for any command's flags.
 
 | Command | Purpose |
 |---|---|
+| [`pipelock guard`](guard.md) | Explain unenforced guard path declarations and immutable compiled-floor refusals. |
 | [`pipelock baseline`](baseline.md) | Inspect, ratify, and relearn behavioral-baseline profiles through the admin API. |
 | [`pipelock demo`](demo.md) | Run self-contained attack scenarios that show what Pipelock catches. |
 | [`pipelock status`](operator.md) | Read-only operator setup and local inspection commands. |
--- a/docs/cli/guard.md
+++ b/docs/cli/guard.md
@@ -1,0 +1,128 @@
+# `pipelock guard`
+
+`pipelock guard` provides read-only inspection of filesystem guard declarations
+and the compiled validation floor.
+
+> **Guard is not enforced in this build.** These commands describe what the
+> configuration declares and what the compiled floor refuses, but no workload
+> is constrained. Both human and JSON output carry this warning.
+
+The commands deliberately separate two policy layers:
+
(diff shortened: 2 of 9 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1179, opened 2026-08-06
feat(assess): sign every assessment, gate content not integrity

luckyPipewrench:
`pipelock assess finalize` put the manifest signature behind the paid `assess` feature, so an unlicensed operator received an assessment they could not verify. On a security product the signature is the part that makes a report mean anything, and charging for it leaves the free report unable to answer the only question that matters about it.

Signing is now the default for every finalized assessment. `--unsigned` remains the explicit opt-out and keeps its meaning for both tiers. First use creates a `pipelock-assess` identity through the shipped keystore helper, so the free path works end to end without hand-wiring key material, and `pipelock assess verify` resolves the same default when no agent is named.

## What stays paid

The `assess` feature continues to gate content: the full assessment JSON and HTML, server-specific findings, remediation commands, exposure analysis, the detailed compliance mapping, and the detached attestation and badge. The free summary keeps its grade, section scores, top findings, redacted server counts, and per-framework compliance coverage counts, and now states its own scope, method identity, signature state, and non-claims.

## Honest signature state

A report claiming to be signed is written before the manifest it commits to can be signed, so any failure between those two points now rolls the claim back: the detached signature is removed, the report is rewritten as unsigned, artifact hashes are refreshed, and the attestation artifacts are purged. Artifact hashing failures are no longer silently discarded, which previously let a signed manifest omit an artifact it should have covered. `--unsigned` also clears a stale signature from an earlier run so file presence cannot contradict the operator's choice.

## Documentation

The README, the tier-gating matrix, the configuration reference, and the compliance guide each described signing as something the `assess` entitlement unlocks. All four now place signing outside the entitlement.

## Summary by CodeRabbit

- **New Features**
  - Finalized assessments are now Ed25519-signed by default, including free summaries.
  - Summaries display signing status, scope, method, compliance coverage, and control counts.
  - Added explicit non-claims information and `--unsigned` support.
  - Verification checks the expected signing key and recorded signer identity.

- **Bug Fixes**
  - Improved handling of invalid, incomplete, or mismatched signing credentials.
  - Failed signing or hashing operations no longer leave misleading signed artifacts.
  - Free summaries limit server identity, while paid reports retain full details.

- **Documentation**
  - Clarified verification, licensing, signing availability, certificate requirements, and sharing guidance.


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1184, opened 2026-08-07
fix(anchor): fsync bundle and marker state before success

luckyPipewrench:
Rooted Unix anchor writes now fsync bundle bytes before rename, then sync the destination directory chain from leaf to receipt root. State-marker creation also syncs the receipt root after the index is written.

Pre-publication sync failures leave no final bundle or temporary file. Post-publication directory-sync failures are returned, and every parent sync is still attempted.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when saving state markers and bundle files by ensuring changes are synchronized to storage.
  - Sync failures are now reported more clearly while remaining synchronization steps continue where possible.
  - Improved cleanup of temporary files when saving encounters an error.
  - Ensured published state remains available even if a later storage synchronization step fails.

- **Tests**
  - Added coverage for synchronization failures, cleanup behavior, publication ordering, and parent-directory synchronization.


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1187, opened 2026-08-07
fix(taint): make MCP action classification deterministic for multi-target calls

luckyPipewrench:
Tool arguments were flattened by ranging a decoded JSON object directly. Go does not specify map iteration order, so an MCP tool call carrying distinct candidate paths or URLs could select a different action reference between runs.

That reference is what an operator's taint trust override is matched against, so the same tool call could honor an override on one run and miss it on the next with no change to the input or the configuration.

The command line the exec path inspects was affected the same way. Its patterns span two tokens, and the walk emits each object key before its value, so an unordered walk could place a value next to a following key and form a pattern that the same input does not form on the next run. That moved identical arguments between protected and elevated sensitivity.

Object keys are now walked in sorted order, so both selections settle the same way every time.

Which candidate wins when a call carries several is unchanged in kind and is tracked separately. Selecting one target and ignoring the rest predates this change and needs a design that can tell an actual target argument apart from an incidental string.

The other JSON walker in this file is left as is. Its consumers are any-match predicates, so iteration order cannot change their result.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when classifying MCP tool calls by ensuring action references are selected in a stable, predictable order.
  * Prevented classification results from varying between runs when processing nested data or shell command arguments.
  * Improved reliability of sensitivity and command classification results across repeated evaluations.

* **Tests**
  * Added coverage to verify deterministic classification across repeated runs and varied input structures.

### DIFF
--- a/internal/mcp/taint.go
+++ b/internal/mcp/taint.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"path"
+	"slices"
 	"strings"
 	"time"
 
@@ -64,6 +65,7 @@ type taintDecision struct {
 	Authority           session.AuthorityKind
 	Result              session.PolicyDecisionResult
 	ActionRef           string
+	OverrideRefs        []string
 	RequiresReauth      bool
 	TaskOverrideApplied bool
 }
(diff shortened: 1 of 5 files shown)


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1188, opened 2026-08-07
fix(recorder): preserve evidence under concurrent writers

luckyPipewrench:
Concurrent recorder processes could interleave a JSON record and its newline or overwrite raw escrow sidecars when they shared a session and sequence. This makes each JSONL record one append, creates escrow sidecars with collision-resistant exclusive names, teaches evidence doctor the new layout, rejects entries that bounded readers cannot consume, and removes incomplete sidecars after failed writes.

## Summary by CodeRabbit

* **New Features**
  * Raw escrow sidecar filenames now include a unique 32-character hexadecimal token.
  * Improved concurrent sidecar creation with collision prevention and bounded retries.
* **Bug Fixes**
  * Incomplete sidecars are cleaned up when writing fails.
  * Recorder writes preserve valid JSONL records and enforce per-entry size limits.
  * Evidence filename parsing supports both legacy and tokenized formats.
  * Escrow sidecars use owner-only file permissions.
* **Tests**
  * Expanded coverage for concurrency, atomic writes, filename validation, cleanup, failure handling, and file rotation.


luckyPipewrench:
@coderabbitai review

DeathCamel58 (approved):

PR #1190, opened 2026-08-07
fix(taint): scan declared MCP command fields instead of a synthetic command line

luckyPipewrench:
Shell classification built the command it inspected by joining every flattened object key and string value, then scanned that synthetic line for multi-token patterns such as git push and sed -i. Object keys are supplied by the caller, so a call could spell a mutating pattern that matched no command the tool would run, and a real command split across fields could be separated so the pattern never formed.

Command content is now taken only from declared command-bearing fields, argument vectors keep their boundaries instead of being glued to unrelated values, and object keys never enter command text. A call with no recognizable command field stays classified as a protected exec rather than being read as benign.

The reported action reference on a call carrying several targets now names every candidate. Trust override matching already keys on the full candidate list, so this changes what an approver and an audit reader see and does not change any enforcement decision. A single-target call reports exactly as it did before.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved shell-command detection to evaluate declared commands and arguments accurately, avoiding false positives caused by unrelated data.
  - Improved classification for shell tools and execution requests, including command vectors, argument arrays, malformed input, and other edge cases.
  - Trust-override results now report accurate action references for single- and multi-target decisions.
  - Multi-target references are deduplicated and displayed together for clearer reporting.

### DIFF
--- a/internal/mcp/taint.go
+++ b/internal/mcp/taint.go
@@ -110,6 +110,7 @@ func evaluateMCPTaint(opts MCPProxyOpts, toolName, argsJSON string) taintDecisio
 	decision.ActionRef = classified.ActionRef
 	decision.ActionRef = mcpActionRef(toolName, decision.ActionRef)
 	decision.OverrideRefs = mcpOverrideRefs(toolName, classified.OverrideRefs, decision.ActionRef)
+	decision.ActionRef = mcpReportedActionRef(decision.OverrideRefs, decision.ActionRef)
 	if tp, ok := opts.Rec.(session.TaskContextProvider); ok {
 		decision.Task = tp.TaskSnapshot()
 		if taintRuntimeTrustOverrideApplies(tp.RuntimeTrustOverrides(), decision.Task, decision.Risk, decision.OverrideRefs) {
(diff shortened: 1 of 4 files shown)


luckyPipewrench:
Both findings reproduce. One is fixed here, one is real but out of scope for this PR, and the severity framing on the first does not hold.

Finding 1, nested command containers, is fixed in c730ef67. A recognized command field holding an object was never walked, so `{"command":{"cmd":"git","argv":["push","origin","main"]}}` reached no matcher and scanned as though no command were present. It now classifies elevated, matching the flat form. The guard was verified by neutralizing the recursive walk and watching its test fail.

The under-enforcement framing on that finding is refuted, and the review flagged it as unproven, so this settles it. `PolicyMatrix.EvaluateWithOptions` returns ask for `ActionClassExec` at normal, elevated and protected sensitivity alike, so the fallback changed reported evidence rather than the taint decision. It was worth fixing as a correctness and evidence defect, not as a weakening.

Finding 2, an argument vector reading as a command, is confirmed and is not fixed here. Two things bound it. The obvious fix is inert: dropping the standalone vector when a base command exists changes nothing, because the combined candidate `echo git push origin main` still contains the substring. That change was written, proven to do nothing, and reverted rather than shipped as a guard that guards nothing. And a real fix has to tell an executable from its arguments, which unanchored substring matching cannot do, while anchoring patterns to the start of a vector would miss chained commands such as `cd /tmp && git push`, which is the wrong direction to fail. That needs a deliberate pass over the matcher and every caller, so it is tracked separately rather than bolted onto this change.

Separately, the review thread about the `program` field is fixed in the same commit. `hasExecIntent` already treated it as an exec while extraction did not recognize it, so its command went unscanned. The two sets are now aligned and a comment says so, since that drift is what caused it.

Celest-Blue (approved):

PR #1195, opened 2026-08-08
fix(runtime): hold required security contracts steady across config reload

luckyPipewrench:
A config reload could turn on `flight_recorder.require_receipts` when no live signed receipt emitter existed. Startup refuses to boot in that state so the misconfiguration surfaces there instead of as an all-blocked proxy at runtime, and the reload path now matches it. The change is ignored, reported on stderr and the audit channel, and a restart with a configured recorder applies it for real.

The reload stages a replacement emitter when the recorder and signing key are already bound and only the emitter is absent or unhealthy. Recovery either installs a healthy emitter and applies the change, or the reload is rejected and the previous posture stays live. Emitter readiness accounts for runtime health rather than startup alone.

Turning a required security contract off through a reload is rejected and the live configuration is preserved. That covers `flight_recorder.require_receipts`, `license_require_intermediate`, `a2a_scanning.require_signed_agent_cards`, `mcp_binary_integrity.require_signature`, `mediation_envelope.verify_inbound.enabled`, and `forward_proxy.sni_require_tls`. Disabling any of them takes a restart.

The rejection compares the old and candidate configurations directly instead of depending on a matching reload warning, so a contract added to that set is covered without a second change in another package.

The evidence corpus integrity alert told operators to stop an evidence export. No export command ships. The remediation now points at `pipelock evidence doctor` and says to treat evidence already published from a damaged corpus as unverified.

Documentation for the forward proxy tunnel settings answered the restart question with "No". Enabling still applies at runtime, disabling takes a restart, and the table says which is which.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened runtime security enforcement by rejecting required-mode downgrades during reloads.
  * Improved receipt enforcement by detecting unhealthy emitters, attempting recovery, and preserving fail-closed settings when recovery fails.
  * Preserved active enforcement settings when applying policy updates.
  * Clarified evidence alerts and directed operators to the supported diagnostic command.

* **Documentation**
  * Clarified restart and hot-reload behavior for SNI/TLS settings and receipt enforcement.


luckyPipewrench:
On finding 1, taking the visibility half and declining the error.

The docs now describe all three outcomes rather than two, and the ignored enable is asserted on the audit channel rather than stderr alone, so a regression that drops the record fails a test.

Returning an error is the part I disagree with. Reload applies one candidate config, so failing it discards every unrelated valid change in the same edit over a setting that needs a restart to take effect either way. That is the availability direction this branch exists to close, and it would trade a narrow reporting gap for a broad refusal.

The automated-control case also does not arise on this path. Conductor apply is the only programmatic caller of Reload, and PreserveConductorBundleLocalRuntimeState assigns the follower's own flight_recorder block over anything a bundle carries, so a bundle cannot set require_receipts at all. The remaining caller is the file reloader, whose operator sees both the stderr warning and the audit event.

Happy to revisit if a programmatic caller that can set this field lands later.

Celest-Blue (approved):

PR #1218, opened 2026-08-10
TEMPORARY PROBE - do not merge - required check binding test

luckyPipewrench:
Temporary diagnostic, opened as a draft and closed as soon as it reports. Do not merge.

It answers one question. The branch ruleset requires a status check whose context is named security-scan. A context is a name rather than a workflow identity, and on a pull_request event GitHub runs the workflow files from the pull request head, so a pull request can define its own job with that name.

This branch deletes one docs file, which makes the real released-binary scan fail, and adds a workflow defining a trivially passing job also named security-scan. If the requirement ends up satisfied, any pull request can mint the required check and the secret-scan gate is advisory rather than enforced. If it stays unsatisfied, the requirement is bound to something a pull request cannot forge.

Run against the repository owner's own repository with their authorization.


luckyPipewrench:
Probe complete, closing. Result recorded for the CI trust work; the branch is deleted and nothing here is intended to merge.

Issue #1186, opened 2026-08-07
Verify external action authority before forwarding mediated requests


Silentpartnercoding:
## Use case

Pipelock already has the right metadata sockets for external action authority: `AuthorityKind`, `AuthorityRef`, actor/auth strength in the mediation envelope, and principal/delegation/target fields in signed receipts. AARP/SVID also deliberately keeps workload identity separate from action authority.

What I could not find is a live path that verifies an externally issued authorization against the exact action before forwarding. Is that a boundary you want Pipelock to support, or is `AuthorityRef` intentionally evidence-only?

A concrete deployment case is an external authority issuing a short-lived authorization for one actor to perform one action against one destination. Pipelock should still independently inspect the traffic, but it should fail closed if the authorization is invalid or does not describe the request it is about to forward.

## Proposed solution

If this fits the intended boundary, add a small provider-neutral verifier interface. Its input would be a normalized action description plus an opaque authority reference; its output would be `allow`, `deny`, or `indeterminate`, with the verified issuer/reference, expiry and machine-readable reason.

The first slice could remain entirely in process, use local test keys, add no network dependency, and cover one maintainer-selected transport. On success, Pipelock would populate the existing authority fields and bind the verified authority reference to the signed receipt. When configured as required, verifier errors and indeterminate results would fail closed.

This would not make Pipelock an identity provider or authorization issuer. It would only verify that an external authorization matches the mediated action at the enforcement boundary.

## Small conformance target

- valid authorization + safe traffic forwards exactly once;
- valid authorization + blocked content forwards zero times;
- missing, expired or revoked authorization forwards zero times;
- actor, action or destination substitution forwards zero times;
- verifier error or timeout forwards zero times;
- the receipt binds the same verified authority reference and action that were enforced.

## Alternatives considered

- Treat `AuthorityRef` as informational metadata: useful for audit, but not live enforcement.
- Treat a verified workload identity as action authorization: proves who the workload is, not whether this action was allowed.
- Add a provider-specific client to core: unnecessary coupling; adapters can remain outside the interface.
- Verify after forwarding: useful evidence, but too late to prevent the effect.

Happy to reduce this further or prepare the in-process reference slice if this matches the intended direction.

luckyPipewrench:
[...]

PR #1225, opened 2026-08-11
feat(authority): add unwired external reference verifier

Silentpartnercoding:
## Summary

Adds the first unwired slice discussed in #1186:

- provider-neutral `Request`, `Result`, and `Verifier` types with a closed `allow` / `deny` / `indeterminate` decision enum;
- a pure in-process Ed25519 reference verifier using local trusted keys and revocation state;
- versioned conformance vectors for exact match, actor/action/destination mismatch, expiry, revocation, malformed input, and timeout.

The local verifier binds a signed reference to the exact request fields it receives. It rejects oversized, malformed, duplicate-key, unknown-field, non-canonical, untrusted, and signature-invalid inputs, and defensively copies its key and revocation configuration.

## Scope boundaries

This is deliberately unwired. It adds no configuration fields and changes no production call sites.

It does not define canonical action construction, gate ordering, redirect re-verification, replay or single-use semantics, or receipt versioning. A verifier `allow` result is only one prospective conjunctive gate: it cannot override an existing scanner, policy, contract, taint, kill-switch, or operator block, and it does not feed or raise internal `AuthorityKind`.

As clarified in the issue, "forwards exactly once" means that one accepted inbound request produces at most one upstream request; it does not mean that a grant is single-use.

## Verification

- `go test -race -count=1 ./internal/authority` on Go 1.25
- `go test -race -count=1 ./internal/authority` on Go 1.26
- `golangci-lint run ./...` with v2.12.2
- 99.1% statement coverage for the new package
- trusted-default-branch `pipelock git scan-diff`: no secrets found

The full repository test command was also attempted locally. Existing macOS-specific `/var` path-resolution, unsupported descriptor-launch, and host-state failures occur in unrelated packages; the new package passes independently under both supported Go versions, and the Linux CI matrix remains authoritative for the draft.

## Summary by CodeRabbit

- **New Features**
  - Added local authorization verification for signed references.
  - Supports trusted issuers, revocations, expiration checks, and request matching.
  - Added standardized allow, deny, and indeterminate decisions with machine-readable reasons.

- **Bug Fixes**
  - Rejects malformed, tampered, expired, revoked, canceled, or unauthorized verification requests.

- **Tests**
  - Added comprehensive coverage for valid, invalid, untrusted, malformed, tampered, and canceled verification scenarios.

### DIFF
--- a/internal/authority/local.go
+++ b/internal/authority/local.go
@@ -1,0 +1,233 @@
+// Copyright 2026 Josh Waldrep
+// SPDX-License-Identifier: Apache-2.0
+
+package authority
+
+import (
+	"bytes"
+	"context"
+	"crypto/ed25519"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/luckyPipewrench/pipelock/internal/jsonscan"
+)
+
+const (
+	localReferencePrefix = "plauth1"
+	localSchemaVersion   = 1
+	maxReferenceBytes    = 16 << 10
+)
+
+var errMalformedReference = errors.New("malformed local authority reference")
+
+// LocalConfig configures an in-process reference verifier. TrustedIssuers and
+// RevokedReferences are copied by NewLocalVerifier. Clock is optional and is
+// intended for deterministic verification and conformance tests.
+type LocalConfig struct {
+	TrustedIssuers    map[string]ed25519.PublicKey
+	RevokedReferences map[string]struct{}
+	Clock             func() time.Time
+}
+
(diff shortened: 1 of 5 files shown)


luckyPipewrench (approved):
Thanks for the great PR! Scoped exactly like we talked about with nothing extra made it easy to review.

One note I'm tracking, nothing for you to change here. The canonical form is whatever Go's `encoding/json` emits, and Go escapes `&` into `\u0026`, so a grant with a normal query string only verifies if the issuer escaped it the same way. I signed one with a raw `&` and a valid signature and got `malformed_reference` back. Fails safe so it's not a hole, but I need to settle that rule before this gets wired.

Thanks again!

PR #1231, opened 2026-08-12
fix(authority): canonicalize local grants with JCS

badmuriss:
## Summary

Use RFC 8785 JCS for local authority grant payloads instead of Go-specific JSON serialization.

This updates the conformance references and adds coverage for a destination containing a literal `&` in its query string.

## Why

This follows the interoperability note raised during the review of #1225: https://github.com/luckyPipewrench/pipelock/pull/1225#pullrequestreview-4910054575

The local verifier previously decoded the payload into a Go struct and compared it with `encoding/json` output. That made Go's field order and HTML escaping part of the signed format. An otherwise valid canonical grant containing a normal query string could fail with `malformed_reference` unless its issuer reproduced Go's `\u0026` escaping.

The repository already has an RFC 8785 implementation in `internal/jcs`. Using it gives local authority references a language-neutral canonical form while preserving fail-closed rejection of duplicate keys, trailing tokens and non-canonical payloads.

This is still an unwired verifier change. It does not add configuration or production call sites. It prepares the reference format for the follow-up work tracked in #1186.

## Testing

- `go test -race -count=1 ./internal/authority`
- `go vet ./internal/authority`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of locally signed payloads using standardized JSON canonicalization.
  * Ensured signed destinations containing ampersands are preserved and accepted correctly.
* **Documentation**
  * Clarified canonical payload handling for local verification.


luckyPipewrench (approved):
Nice, thanks!

PR #1245, opened 2026-08-13
fix(doctor): report action divergence under its own check name

luckyPipewrench:
`pipelock doctor` and `pipelock check` reported sub-detector action divergence under `config_exemption_semantics`, so an operator whose exemptions and suppressions were correct saw an exemption warning and had no reason to look at enforcement strength. Action divergence now reports under `config_action_divergence`.

Check names came from a two-branch default: the suppress scope got `config_suppress_semantics` and every other scope fell through to `config_exemption_semantics`. That held while every other scope described an exemption. Scopes now route through an explicit mapping, and a test pins the whole table, so a scope that is not an exemption fails here instead of reaching an operator as a mislabeled warning.

A second change keeps the suppress and exemption surface visible. The ok check that reports a clean surface appeared only when nothing else had a finding, so an unrelated advisory removed that surface from the report and left no way to tell a clean suppression from a check that never ran. It now appears whenever the suppress and exemption analysis produces no findings of its own.

`docs/cli/doctor.md` lists the new check.

## Summary by CodeRabbit

* **New Features**
  * Added a diagnostic check for sub-detectors configured with weaker actions than their enclosing section.
  * Action-divergence findings now appear under a dedicated check name.

* **Bug Fixes**
  * Suppression and exemption results are reported correctly alongside unrelated advisories.
  * Added an explicit successful result when no suppression or exemption issues are detected.
  * Unrecognized diagnostic scopes no longer appear as exemption findings.

* **Documentation**
  * Documented the new action-divergence diagnostic check and its advisory behavior.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1247, opened 2026-08-13
fix(cli): refuse outputs that name the signing key they just used

luckyPipewrench:
## What changed

Commands that sign or verify and then write a result now refuse when the output path is the same file as a key they loaded. Coverage certificates, fleet reports, MCP integrity signatures, learn compile and lifecycle receipts, verify-receipt clean reports, and local anchor bundles are covered. Hex `--key` values are not treated as paths.

`airlock.triggers.on_severity` and `anomaly_count` no longer load. Those fields validated and never fired. Rule bundles that set `pattern.exempt_domains` still ignore the field, because bundles are deny-only, and now warn at load instead of staying silent.

Failure direction: each new refusal fails closed, so a command that would have overwritten a loaded key now stops rather than proceeding. The fields that no longer load were inert before this change, so nothing that previously enforced stops enforcing. What a reviewer should distrust most is the path comparison itself, since a key referenced by a different but equivalent path would not be caught by a naive string match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,27 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Breaking Changes / Upgrade Notes
-
-- **ACTION REQUIRED — `airlock.triggers.on_severity` and a nonzero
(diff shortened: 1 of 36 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1261, opened 2026-08-15
fix(mcp): fail closed at capacity bounds and require signed resets

luckyPipewrench:
## What changed

Capacity limits across MCP tool inventory, tool drift baselines, A2A agent-card baselines, cross-request fragment buffers, the envelope replay cache, and authenticated listener state now deny and report when reached. Each one used to drop the tracking state it needed and keep going, so a control could stop applying while every surface still described it as enabled. Reaching a bound now produces a refusal with a named reason and a metric, matching the tool-chain session-capacity path that already behaved this way.

Reset authority for drift re-baselining and the adaptive airlock no longer rests on who owns a file. An operator mints a short-lived delegation signed with a dedicated key, bound to reset kind, target, proxy instance, current epoch, expiry, and a single-use nonce, and the proxy holds only the public half. Where an agent and the proxy share a user account, file ownership proved nothing about who asked for the reset, so this replaces that mechanism rather than tightening it. A delegation is a one-shot capability: someone who obtains a valid one can spend it once before it expires, and cannot forge another.

Sandboxed MCP startup keeps working alongside the proxy hardening it collided with. A mapped launch stays readable only long enough to write the child's ID map, the child waits on a parent-held gate, and the target starts only after the parent has hardened. Hardening or release failure kills and reaps the child rather than proceeding.

Linux process limits no longer apply a fixed per-user process ceiling that may already be below the host user's live task count. Every sandbox mode now sets that ceiling from current shared-user usage plus 1,024 tasks. This keeps a fork bomb bounded without making a busy host abort the protected child before its target starts.

A configuration reload that turns on drift detection no longer clears the existing baseline. Re-baselining is an operator action and now requires a delegation on every path that reaches it.

The response pattern for instruction-override text now requires an authority or action context near the phrase. It matched ordinary prose inside shipped blocking presets, and a first attempt at narrowing it over-corrected by depending on one separator character.

### Breaking change

The reset control-file paths are unchanged, but their contents are not. `listener_drift_reset_file` and `--adaptive-reset-file` now require a signed delegation, so an existing marker file is rejected as malformed. Operators generate a reset-authority key, export its public half into configuration, and mint a delegation for each reset. The full workflow, including rotation, revocation and recovery, is in `docs/mcp/per-server-false-positives.md`.

### Failure direction and what to distrust

These bounds refuse where they previously continued, so the failure direction moved from allowing unverified traffic to refusing legitimate traffic under load. The limits are shared across clients on a single proxy and there is no per-client accounting yet, so read the availability direction of each new refusal before the security direction.

Process-environment hardening is load-bearing on two paths only: an unsandboxed proxy, and the best-effort sandbox fallback that runs without user namespaces. Under full or strict namespaces the user-namespace boundary already denies that read, so a passing result there does not on its own demonstrate the hardening works.

The Linux process limit is still shared-user accounting rather than a child-owned cgroup limit. Other processes under the same user can consume its headroom, so treat it as a bounded fallback with an availability tradeoff, not isolated per-sandbox accounting.

## Summary by CodeRabbit

* **New Features**
  * Added signed, single-use authorization workflows for MCP resets, including creation, inspection, public-key export, and revocation.
  * Dashboard receipt links now use globally unique action IDs.
  * Added clearer capacity metrics and reporting for blocked requests and dropped records.

* **Bug Fixes**
  * Improved detection of directive-based “New Instructions” content while reducing benign matches.
  * Capacity limits now fail closed without evicting trusted state or partially accepting data.
  * Strengthened sandbox, subprocess, and replay-cache protections.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1269, opened 2026-08-15
fix(sandbox): preserve capacity on busy shared users

luckyPipewrench:
## What changed
- Keep sandbox launches available on busy shared-user hosts by granting up to 1,024 tasks of headroom without exceeding the absolute 4,096-task ceiling.
- Ignore transient and foreign process-status entries while counting the sandbox user's tasks; launch still fails closed at the ceiling or on malformed accounting for that user. Reviewers should distrust any path that can set the limit below current usage or above the ceiling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux process-limit handling so requests below the configured ceiling are accepted reliably.
  * Requests exceeding available capacity are safely capped instead of failing unnecessarily.
  * Requests at or above the ceiling are rejected consistently.
  * Process and thread counts now handle empty, incomplete, or unrelated system entries more reliably.
  * Added safeguards for maximum-value and live-system scenarios to prevent overflow-related issues.

### DIFF
--- a/internal/sandbox/rlimit_linux.go
+++ b/internal/sandbox/rlimit_linux.go
@@ -22,7 +22,7 @@ import (
 // These prevent fork bombs, disk fill, FD exhaustion, and core dumps.
 const (
 	rlimitNProc           uint64 = 4096    // absolute shared-UID task ceiling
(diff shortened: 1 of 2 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1285, opened 2026-08-18
fix(sandbox): bound bridge relays with idle timeout and shutdown cancellation

glatinone:
## Summary

- bound sandbox bridge relays with a shared bidirectional idle timeout
- close active relays when bridge context cancellation or shutdown occurs
- preserve TCP/Unix half-close semantics for normal EOF
- handle short writes without silently dropping bytes
- add regression coverage for stalled connections, positive activity refresh in both directions, client-write half-close, and cancellation/close races

## Follow-up remediation

The bridge timeout is currently initialized to the forward proxy default and can be adjusted through `BridgeProxy.SetIdleTimeout`. The remaining review item is to transport the resolved `forward_proxy.idle_timeout_seconds` value into both sandbox child launch paths.

Signed-off commits use `Kiell Tampubolon <yeheskieltampubolon@gmail.com>`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable idle timeouts for bridge connections, including runtime adjustment and the option to disable them.
  * Propagated bridge timeout settings through sandbox and standalone launch configurations.
  * Context cancellation now fully closes active bridge connections.

* **Bug Fixes**
  * Improved shutdown reliability for stalled or failed transfers.
  * Activity in either direction now keeps connections alive.
  * Preserved half-close behavior during normal connection termination.


luckyPipewrench:
@glatinone, are you still planning to finish this PR? The CLA is still unsigned and the two CodeRabbit threads have been open since August 18.

I checked both findings against the current head and they're valid. `SetIdleTimeout` has no production caller, so both sandbox launch paths still use 120 seconds even when an operator configured a different `forward_proxy.idle_timeout_seconds`. The tests also don't cover the two behaviors the new relay depends on: activity keeping a live connection open, and one side reaching EOF while the other direction still works.

The title also says this adds a kill-switch check, but I can't find any kill-switch wiring in the patch. The new stop path handles context cancellation and bridge shutdown. Please either add the missing kill-switch behavior or rename the PR to match what it does.

If you're continuing with it, please sign the CLA, rebase onto current `main`, fix and resolve both review threads, and update the title and description. If you're not planning to continue, say so and I'll close it.

PR #1331, opened 2026-08-23
fix(evidence): stream legacy compaction safely

luckyPipewrench:
## What changed
- Verify historical recorder entries from their exact stored detail bytes, then stream offline compaction without a whole-session memory ceiling while keeping normal evidence reads bounded.
- Verify recorder checkpoints and both receipt generations before and after staging, and fail closed on malformed input, unknown record types, source changes, or chain disagreement. Reviewers should distrust any path that can publish staged shards without proving byte-for-byte source equivalence and signed-chain parity.

## Summary by CodeRabbit

* **New Features**
  * Added safer offline evidence compaction with bounded streaming, shard limits, integrity verification, and atomic archival.
  * Added support for verifying legacy and current receipt formats before and after compaction.
  * Improved handling of large shards while preserving metadata and verification details.

* **Bug Fixes**
  * Malformed records can no longer produce signed contracts.
  * Resume operations now detect tampered or inconsistent evidence.
  * Raw evidence details are preserved consistently for hash verification.

* **Documentation**
  * Documented offline legacy-evidence compaction and recovery procedures.

### DIFF
--- a/docs/guides/receipt-verification.md
+++ b/docs/guides/receipt-verification.md
@@ -657,6 +657,25 @@ pipelock verify-receipt receipt.json --key /etc/pipelock/keys/flight-recorder-si
 The exported file is public key material only. Do not hand verifiers the private
 file named by `flight_recorder.signing_key_path`.
 
+## Offline compaction of legacy evidence
+
+Online evidence views deliberately stop at their documented per-shard read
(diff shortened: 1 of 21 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1332, opened 2026-08-23
fix(contain): disable npm lifecycle scripts by default

luckyPipewrench:
## What changed
- Set npm lifecycle-script suppression in both the managed user config and contained runtime environment, preventing project config from silently re-enabling install scripts.
- Document the per-command override and reinstall requirement; reviewers should distrust any launch path that omits the runtime contract.

## Summary by CodeRabbit

* **New Features**
  * npm lifecycle scripts are now disabled by default in managed environments.
  * Explicit command-line overrides remain available for trusted dependencies.
  * Reinstalling applies the updated runtime settings to existing installations; upgrades leave agent tool configuration unchanged.
  * Agent environments now consistently apply the lifecycle-script restriction across supported shells and npm configuration.

* **Documentation**
  * Updated runtime contract documentation to describe lifecycle script controls and configuration behavior.


luckyPipewrench:
Both findings correctly observe that npm install --ignore-scripts=false overrides the default. That is intentional and documented. This setting is defense in depth against dependency lifecycle scripts running automatically; it is not the containment boundary or an attempt to prevent an agent that already executes arbitrary commands from opting out. The enforced boundary remains the contained user’s nftables egress policy, which the npm override does not change. A live npm precedence check confirmed both the default suppression and the explicit override. No code change is needed.

Celest-Blue (approved):

PR #1334, opened 2026-08-24
fix(proxy): enforce session policy on redirects

luckyPipewrench:
## What changed
Re-evaluate taint policy and scoped airlock state before each fetch or forward redirect, failing closed when the redirected hop is denied. Block receipts preserve the redirect-time decision; distrust any path that reports only the admitted request's state.

## Summary by CodeRabbit

* **Security**
  * Redirected requests now re-evaluate taint and access policies at every hop.
  * Protected actions can be blocked, require explicit approval, or be restricted by airlock controls before redirected traffic is sent.
  * Blocked-request records now reflect the taint details from the redirect that triggered the block.

* **Reliability**
  * Session-based policy decisions and approvals are consistently preserved across redirects.
  * Harmless redirects continue without unnecessary interruption across supported request flows.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1343, opened 2026-08-25
fix(mcp): enforce A2A card controls on proxy responses

luckyPipewrench:
## What changed

Agent Card signature verification and drift enforcement now run on MCP proxy responses. They were fully implemented and tested but had no production caller, so an operator who enabled them in blocking mode got no enforcement on any MCP transport while the same checks worked on the forward-proxy and TLS-intercept paths.

The wiring goes into the shared response forwarder that every MCP transport converges on, rather than into one transport, so the parity is structural. Card routing no longer depends on an exact method-name match either: a card-shaped result, or a card method whose name differs in case, now takes the Agent Card path instead of falling through to generic scanning.

The failure direction to distrust is the fall-through. When A2A scanning is disabled, or a response is ordinary MCP traffic, this code must reach the generic scanner with the per-server suppression context intact, and it must not start blocking traffic that previously passed. Reviewers should confirm the disabled and non-A2A paths behave exactly as before, and that an undetermined card shape does not silently take the permissive branch.

## Summary by CodeRabbit

* **Security Enhancements**
  * Applied consistent Agent Card signature enforcement across HTTP, WebSocket, and subprocess MCP proxy transports.
  * Unsigned or changed Agent Cards can be blocked when verification is enabled.
  * Signed Agent Cards continue to pass validation with authorization scoped to the appropriate origin.

* **Reliability**
  * Improved handling of authenticated card methods, malformed JSON-RPC messages, batches, and invalid card payloads.
  * Preserved existing response filtering for ordinary MCP traffic.

### DIFF
--- a/internal/cli/runtime/mcp.go
+++ b/internal/cli/runtime/mcp.go
@@ -259,6 +259,17 @@ func handleProxyError(err error, logW io.Writer, sentryClient *plsentry.Client) 
 	return err
 }
 
+func applyMCPA2AOpts(opts *mcp.MCPProxyOpts, cfg *config.Config, baseline *mcp.CardBaseline, cardURL string) {
+	if cfg != nil {
+		a2a := cfg.A2AScanning
+		opts.A2ACfg = &a2a
+	}
+	opts.CardBaseline = baseline
+	if cardURL != "" {
+		opts.A2ACardURL = cardURL
+	}
+}
+
 func applyMCPResponseSuppressOpts(opts *mcp.MCPProxyOpts, cfg *config.Config, serverName string) {
 	opts.ServerName = serverName
 	trust := config.ResponseTrustUntrusted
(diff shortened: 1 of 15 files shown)


luckyPipewrench:
Rejecting the medium finding on `internal/mcp/proxy_ws.go:52`, with the reproduction below rather than on authority.

The observation is correct: `RunWSProxy` sets `A2ACardURL` and never sets `A2ACardAuthFingerprint`, while the HTTP forward path derives one at `internal/mcp/mcp_http_forward.go:140`. The premise built on it is not. The finding assumes responses on a WebSocket connection can arrive under different upstream credentials, and that path has no upstream credential at all.

Three structural facts, each checkable at `6443dbc3`:

The WebSocket transport constructors accept only a URL. `NewWSClient(ctx, rawURL)` and `NewWSClientWithDialer(ctx, rawURL, dialContext)` in `internal/mcp/transport/wsclient.go` take no header argument, so the WebSocket upstream cannot carry an Authorization header even if one were configured.

`UpstreamHeaders` is documented as HTTP-only at `internal/mcp/opts.go:122` and is consumed exclusively in `internal/mcp/mcp_http_reverse.go`. `proxy_ws.go` contains no reference to it, nor to any header.

The only production assignment of `UpstreamHeaders` is onto the listener options in `internal/cli/runtime/mcp.go:1370`, which is the HTTP listener path rather than the WebSocket one.

So there is no set of differing upstream credentials on this transport for a drift baseline to be partitioned across. Populating `A2ACardAuthFingerprint` there would record a constant empty fingerprint, which is already the value it holds, and would suggest a partitioning guarantee the transport cannot provide.

Leaving this open for rebuttal. If WebSocket upstream credentials are added later, the fingerprint has to be derived at the same time, and that is worth a comment at the call site so the omission is not read as an oversight a second time.

Celest-Blue (approved):

PR #1388, opened 2026-08-29
fix: refuse unsafe installs and anchor posture evidence

luckyPipewrench:
## What changed

- Receipt-enabled runtimes record missing or unreadable posture proofs in unsigned `session_open.ext` metadata. Required containment still refuses startup unless a pinned signer and verified capsule prove containment.
- Real `contain install` runs stop before host setup in recognized container or init-image environments and on non-Linux systems. `--dry-run` still prints the plan.

The advisory field doesn't change signed v1 canonical data. Receipt-chain hashes cover the full envelope, so changing the field after recording breaks the next chain link.

## Summary by CodeRabbit

- **New Features**
  - Added flight recorder options to require verified containment evidence and configure a trusted posture signer key.
  - Signed session receipts now report posture proof availability (`attested`, `absent`, `unreadable`, or `invalid`).
  - Containment installation now validates the host environment and prevents unsupported installations.

- **Bug Fixes**
  - Invalid or insufficient containment evidence now fails closed when required.
  - Dry-run installations continue without host preflight checks.

- **Documentation**
  - Documented configuration prerequisites, evidence statuses, startup behavior, and restart-only reload changes.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2666,6 +2666,8 @@ flight_recorder:
   retention_days: 90
   redact: true
   require_receipts: false
+  require_containment_evidence: false
+  posture_signer_key: "/path/to/posture-signer.pub"
   sign_checkpoints: true
   signing_key_path: "/path/to/signing-key"
   max_entries_per_file: 10000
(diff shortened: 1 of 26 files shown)


luckyPipewrench:
Splitting this. The startup crash fix is now #1390, with no change to the signed receipt surface.

Review raised that adding `posture_availability` to the frozen v1 signing projection makes new receipts unverifiable by already-installed verifiers, and that the field carries diagnosis rather than a containment invariant, since an absent binding already cannot be upgraded from a separately supplied capsule. One of its four states also cannot be emitted, because an invalid proof aborts before any receipt is written.

The typed availability work, the pinned posture signer key, and the enum validation stay on this branch and will argue for themselves once the shape is settled.

Celest-Blue (approved):

PR #1399, opened 2026-08-30
fix(conductor): scope admin requests to the authenticated organization

luckyPipewrench:
## What changed

The Conductor admin authorizer was constructed without an organization, so an admin credential authenticated for one org could act on any org. The publish authorizer built a few lines below it already carried both org and fleet scope, which made the gap easy to miss: the two sat adjacent with opposite scoping, and the admin constructor's name said scoped while the credential passed to it was not.

Authentication now returns the authenticated org and fleet identity, and every administrative operation proves its target falls inside that scope. This covers the whole class rather than the two operations that prompted the review: remote kill, rollback, rollback clear, both dry-run paths, enrollment create, list and revoke, and follower removal. A cross-scope clear or revoke returns not-found without mutating anything, so the response cannot be used to enumerate objects in another org.

This is a deliberate breaking change in two directions. The exported authorizer changes shape, and an administrative request that previously reached another organization now fails. A compatibility shim was considered and rejected because it would restore the gap it exists to close. No configuration change is required, since the admin org flag was already mandatory.

## Summary by CodeRabbit

* **Security Enhancements**
  * Admin credentials now enforce organization and fleet scope for emergency controls, enrollment tokens, follower management, and fleet-skew overrides.
  * Requests targeting resources outside an administrator’s permitted scope are rejected.
  * Cross-scope rollback and enrollment-token operations return appropriate not-found or forbidden responses without modifying data.

* **Bug Fixes**
  * Improved rollback authorization lookup and handling for missing or inaccessible records.
  * Rollback records cannot be cleared when their stored authorization has been replaced.
  * Enrollment-token filtering now consistently respects administrator scope.

### DIFF
--- a/enterprise/cli/conductor/cmd.go
+++ b/enterprise/cli/conductor/cmd.go
@@ -157,8 +157,8 @@ func serveCmd() *cobra.Command {
 	cmd.Flags().StringVar(&opts.adminTokenFile, "admin-token-file", "", "file containing bearer token required for Conductor admin requests")
 	cmd.Flags().StringVar(&opts.auditorOrgID, "auditor-org", "", "org scope for the auditor bearer token on audit/follower read requests (required)")
 	cmd.Flags().StringVar(&opts.auditorFleetID, "auditor-fleet", "", "optional fleet scope for the auditor bearer token on audit/follower read requests")
-	cmd.Flags().StringVar(&opts.adminOrgID, "admin-org", "", "org scope for the admin bearer token on audit/follower read requests (required)")
-	cmd.Flags().StringVar(&opts.adminFleetID, "admin-fleet", "", "optional fleet scope for the admin bearer token on audit/follower read requests")
+	cmd.Flags().StringVar(&opts.adminOrgID, "admin-org", "", "org scope for every admin bearer request (required)")
+	cmd.Flags().StringVar(&opts.adminFleetID, "admin-fleet", "", "optional fleet scope for every admin bearer request")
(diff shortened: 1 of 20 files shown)


luckyPipewrench:
Rejecting the cross-org clear finding by reproduction, and fixing the other four.

The subtest does construct a genuinely cross-organization case. `signedTestRollback` builds records with `OrgID: "test-org"`, and the test admin authenticates as `org-main`/`prod`, so the record is out of scope and the not-found branch is reached for the reason the test claims.

Proven rather than argued. Removing the scope half of the guard, leaving `if !found {`, makes that exact subtest fail with `cross-org clear status=200 body={"authorization_id":"rollback-other-org","cleared":true}`. A test that did not exercise a cross-org record would have stayed green.

The four CodeRabbit findings were all real and are fixed in `c8339ab0e`. The rollback clear read a record, checked its scope, then deleted through a separate call, so an id remapped in between would have had the replacement deleted by an administrator authorized for a different record. The clear is now bound to the authorization hash and the store re-checks that binding under its own write lock, reporting a mismatch identically to an absent record. A store that cannot express the binding causes a refusal rather than an unconditional delete.

Also normalized the revoke token id so the scope lookup and the revoke agree on which token was named, added the fleet-denial and organization-wide branches of the scope check, and changed both cross-scope denial tests to prove the foreign record is unchanged rather than only asserting the status code.

Worth noting one of my own regression tests was vacuous on the first attempt: the double did not satisfy the base clearer interface, so the handler returned before reaching the bound path and every assertion passed against untouched fields. That is recorded in a comment on the test.

Celest-Blue (approved):

PR #1413, opened 2026-08-30
perf(scanner): gate DLP patterns on required literals

luckyPipewrench:
## What changed
- Parse compiled DLP regex syntax at scanner construction and index only required ASCII literals proven across every match path. Patterns with any unprovable branch stay in the always-run set, so uncertainty runs the regex instead of skipping detection.
- Deduplicate candidates when one input contains multiple anchors for the same pattern. Existing clean-URL benchmarks improve by about 12% to 36% with allocation counts unchanged, and the tests cover configured plus immutable-core patterns.

## Summary by CodeRabbit

* **Performance**
  * Improved DLP scanning by identifying multiple safe text anchors within detection patterns.
  * Reduced unnecessary pattern evaluations while preserving coverage for patterns without reliable anchors.
  * Eliminated duplicate candidate evaluations during scanning.

* **Bug Fixes**
  * Improved handling of alternations, repetitions, character classes, and case variations.
  * Ensured generic credential patterns continue to be evaluated conservatively.

* **Tests**
  * Added comprehensive coverage for anchor extraction, fallback behavior, candidate selection, and randomized pattern matching.

### DIFF
--- a/internal/scanner/prefilter.go
+++ b/internal/scanner/prefilter.go
@@ -4,6 +4,7 @@
 package scanner
 
 import (
+	"regexp/syntax"
 	"sort"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1417, opened 2026-08-31
fix(scanner): preserve response regex alternation branches

luckyPipewrench:
## What changed

- Derive response prefilter anchors from the shared regex syntax tree so every branch contributes a required literal before the scanner can skip a pattern. Patterns without a proven anchor still run.
- Cover configured and immutable response patterns across each normalization pass, and keep candidate ordering deterministic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pattern matching for response scanning, including nested optional groups, leading anchors, and alternation branches.
  * Ensured relevant patterns are consistently selected during pre-filtering, with safe fallback behavior for unsupported or non-ASCII patterns.
  * Made pattern evaluation order deterministic.

* **Tests**
  * Expanded coverage for response filtering, regex edge cases, always-run behavior, and generated matching samples.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1419, opened 2026-08-31
fail closed on missed request, media, and receipt identity

luckyPipewrench:
## What changed

The HTTP proxy refuses request trailers it can't scan, sends the Host that policy already admitted, and won't replay a scanned body onto a different host after a redirect.

MCP media policy now inspects nested resource blobs. Strict reload rejects dropping the MCP listener state-token requirement. Pinned receipt verify binds signer_key_id to the pinned key.

Missing those fields used to look like a clean allow or a clean verify. The ignored field still went through.

## Summary by CodeRabbit

- **Security Enhancements**
  - Pinned-key receipt verification now rejects relabeled or mismatched signer identities across supported verifiers.
  - HTTP trailers are scanned safely; requests containing unscanned trailer data are blocked.
  - Redirects that replay request bodies to a different destination are blocked.
  - Client-controlled `Host` values can no longer override the approved destination.

- **Media Protection**
  - MCP media policies now inspect embedded resource audio, video, and image content.

- **Configuration**
  - Strict-mode reloads reject disabling listener state-token requirements and provide clearer warnings.

### DIFF
--- a/cmd/pipelock-verifier/evidence.go
+++ b/cmd/pipelock-verifier/evidence.go
@@ -153,7 +153,7 @@ func verifyEvidenceReceipt(r contractreceipt.EvidenceReceipt, keyHex string, opt
 		}
 		return false, nil
 	}
-	if err := contractreceipt.VerifyWithKey(r, chainOpts.PinnedKey, r.Signature.SignerKeyID); err != nil {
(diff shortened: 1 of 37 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1446, opened 2026-09-03
fix(scanner): share injection regex, framed directives still block

luckyPipewrench:
## What changed

The core prompt-injection regex now lives in one constant shared by the immutable core floor and the default response-scanning pattern, and a parity test fails if the core copy, the default config, or any shipped preset drifts from it.

A regression corpus pins down that untrusted content can't exempt itself from the core match by dressing a directive as documentation. Quotes, table cells, fenced examples, code spans, curly or CJK quotation marks, zero-width characters, and a directive placed after its own framing all still match on the original, normalized, and reassembled views, through both the core and configured paths. Security documentation that quotes an injection string keeps blocking. Letting a specific documentation host through remains an operator decision, not something the page can claim for itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency of prompt-injection detection across core scanning and configured scanner presets.
  * Confirmed that documentation-framed, normalized, and reformatted prompt-injection phrases continue to be blocked.
  * Added safeguards to keep prompt-injection matching aligned across scanning configurations.
  * Improved example verification reliability by handling temporary upstream package timeouts gracefully and skipping affected checks instead of reporting misleading failures.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1451, opened 2026-09-04
fix(scanner): refuse exempt_domains on core DLP pattern names

luckyPipewrench:
## What changed

A custom DLP pattern that reuses a core safety-floor name (`AWS Access ID`, `Private Key Header`, and the other core names) can no longer carry `exempt_domains`. Config validation rejects it at startup and on hot reload, and the URL scanner's per-view and query-subsequence paths ignore the field for core names even when no validation ran, matching the guard the body and response filters already have.

The core floor already blocks these credentials first, so the visible change is that an inert exemption now fails loud instead of reading as an allowance and the configured layer can no longer disagree with the floor. Custom pattern exemptions are unchanged, and no shipped preset sets one on a core name.

## Summary by CodeRabbit

- **Bug Fixes**
  - Core safety-floor DLP patterns can no longer be exempted for specific domains.
  - Invalid exemptions are rejected during configuration startup and reload.
  - Core credential protections remain enforced across all destinations, including query scanning.
  - Custom DLP patterns continue to support domain exemptions.

- **Documentation**
  - Clarified the rules for per-pattern domain exemptions.
  - Updated examples to use configurable patterns when demonstrating domain exemptions.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -785,6 +785,8 @@ dlp:
 
 For built-in provider-key patterns, the default config already exempts the provider's own API host for URL DLP and adds matching `suppress` entries for request-body and request-header DLP. The same key is still blocked when sent to any other destination. See [Provider-Key DLP Coverage](security/provider-key-dlp-coverage.md) for included shapes, exclusions, and the custom provider-key path.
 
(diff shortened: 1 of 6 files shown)


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1453, opened 2026-09-04
fix(proxy): apply SigV4 credential routes on the reverse proxy

luckyPipewrench:
## What changed

The reverse proxy now applies `request_body_scanning.sigv4_credential_routes` the same way the forward proxy and intercepted CONNECT already do, so a presigned URL inside a body on a configured route isn't blocked by the DLP floor there anymore. Out-of-route hosts still block.

WebSocket frames deliberately stay outside the routes. A frame has no request method or declared content type for an exact route to match, and the frame path already keeps HTTP entropy route exceptions off for the same reason. That refusal is now explicit and tested, and the docs state the transport scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Applied SigV4 credential-route settings during HTTP request body scanning, including requests behind an upstream base path.
  * Ensured route-based exceptions do not apply to WebSocket frames, so embedded AWS access keys remain blocked.
  * Clarified that HTTP exceptions require matching host, path, method, content type, AWS-owned hosts, and complete SigV4 URL structure.

* **Documentation**
  * Updated configuration guidance for `sigv4_credential_routes` behavior across HTTP and WebSocket traffic.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1454, opened 2026-09-04
fix(proxy): give request-body trust its own list

luckyPipewrench:
## What changed

Request-body scanning gets its own `trusted_hosts` list. On a listed destination, injection-shaped text in a request body and a critical credential that redaction fully rewrote follow the configured action instead of the hard block; every scan still runs and other findings keep their actions.

Before this the request side read `response_scanning.exempt_domains` for those two decisions, while the docs said that list only skips the response injection scan and never loosens outbound controls. The docs were the contract, so the request side now consults only its own list, and a host exempted for responses no longer changes what leaves the agent. No shipped preset sets response exemptions, so defaults don't change; a hand-edited config that relied on the old coupling sees the documented hard blocks until the host is added to `request_body_scanning.trusted_hosts`. The canonical policy hash goldens moved because the new field is policy-bearing.

## Summary by CodeRabbit

- **New Features**
  - Added `request_body_scanning.trusted_hosts`, allowing configured actions—such as warnings—to apply to trusted destinations while request-body scanning remains active.
  - Supports exact and wildcard host patterns with case-insensitive matching.

- **Bug Fixes**
  - Response-scanning exemptions no longer weaken request-side protections for prompt injection or fully redacted critical DLP findings.
  - Trusted-host entries are validated consistently, including when request-body scanning is disabled.

- **Documentation**
  - Clarified trusted-host behavior and configuration in the configuration and redaction guides.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1486, opened 2026-09-06
feat(proxy): return the recorded receipt id to the caller

luckyPipewrench:
## What changed

Every mediated HTTP response that a signed receipt covers now carries `X-Pipelock-Receipt` with that receipt's `action_id`, so a caller can pair its own request id with the proxy's evidence and later prove which receipt covered the request. The header means a receipt was recorded before the response was written. It's set only after emission succeeds, never from an id that was allocated but not recorded, and a proxy running without a receipt emitter sends no header at all.

- Blocks on every transport set it, and allow responses set it only under `flight_recorder.require_receipts`, where emission happens before egress. Best-effort allow responses stay uncorrelated on purpose; the guide says so. Streaming reverse-proxy findings that arrive after the response is committed keep their receipt but can't carry the header.
- Any `X-Pipelock-Receipt` value an upstream sends is removed on every response path, so the only value a caller can see is one the proxy minted. The signed receipt format, the replay publication gate and the other-language verifiers are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added the `X-Pipelock-Receipt` response header with the action ID of a receipt successfully recorded by the proxy.
  * Available for blocked requests and eligible allow responses when receipt recording occurs before forwarding.
  * Supports request correlation; callers must verify the signed receipt before relying on it.
* **Bug Fixes**
  * Prevented unrecorded or upstream-supplied receipt headers from being presented as proxy receipts.
* **Documentation**
  * Documented header behavior, limitations, and browser access requirements.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

socksninja (commented):
This is close to the commercial boundary I’m testing. `X-Pipelock-Receipt` already lets a caller correlate a mediated request with a successfully recorded receipt. The remaining question is narrower: can one real agent execution bind that receipt to an independently rereadable external outcome, rather than only to the proxy-observed action?

A minimal control experiment: one harmless mediated tool/action; capture the agent `run_id` or equivalent, the emitted receipt `action_id`, then from a fresh process reread the external state and hash the observed state. If the chain is already reconstructible, I’d treat that as reusable evidence. If not, the exact missing link becomes a concrete reliability/governance gap worth testing with an operator.

No SABLE integration or protocol change requested.

socksninja (commented):
One commercial test follows directly from this boundary. PipeLab now sells enterprise fleet governance and signed evidence, so the question is not whether evidence matters; it is whether an operator will pay to close the remaining gap between “the mediated action was authorized/recorded” and “the intended external outcome was independently verified.”

I’d like to run one bounded design-partner experiment: one harmless real action, one Pipelock receipt, one fresh-process external readback, and one verifier-side state hash. Success is a portable record a third party can verify without trusting the agent runtime. If that closes a real audit/governance decision for a customer, we have a concrete paid-pilot boundary; if it does not, we cut this wedge.

I’m intentionally not asking for a protocol change or SABLE integration—just a buyer-facing experiment.

PR #1495, opened 2026-09-07
fix(gauntlet): authenticate the pinned release lookup

luckyPipewrench:
## What changed

The scheduled Gauntlet resolves its pinned Pipelock release through the GitHub API without credentials, so the request draws on a per-IP rate limit shared with every other job on the hosted runner.
The 2026-09-07 scheduled run returned 403 at that call and produced no score at all.

The benchmark step now passes the workflow token, which moves the lookup onto a per-repository budget.
The runner still functions without a token, so an exhausted budget fails the run rather than reporting an empty result as success.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the reliability of scheduled benchmark runs by using authenticated access for GitHub release lookups, reducing failures caused by API rate limits.

### DIFF
--- a/.github/workflows/continuous-gauntlet.yaml
+++ b/.github/workflows/continuous-gauntlet.yaml
@@ -91,6 +91,16 @@ jobs:
 
       - name: Run portable canonical benchmark
         shell: bash
+        # The runner resolves the pinned Pipelock release through the GitHub
+        # API. Unauthenticated calls are rate-limited per source IP, which on a
+        # hosted runner is shared with every other job on that machine, so the
+        # lookup returns 403 and the whole scheduled run produces no score. The
+        # run on 2026-09-07 failed exactly this way. GITHUB_TOKEN raises the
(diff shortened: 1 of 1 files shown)


luckyPipewrench:
Closing unmerged. The 403 this addressed is real, but authenticating that call is the wrong fix.

`scripts/test_gauntlet_candidate_workflow.py` asserts this workflow carries no credential, alongside assertions that it cannot open pull requests, reach the site, or run the promotion script. The invariant is that the job producing benchmark scores is structurally unable to publish one, and that is worth keeping.

The step also runs the benchmark runner checked out from another repository, so passing a token would hand a credential to code this repository does not own.

The durable repair belongs in the benchmark runner, which owns the release lookup: bound retry against the rate-limited metadata call, escalating to a recorded fallback that relies on the pinned checksum for artifact identity only if retries prove insufficient. Tracked for that repository instead.

PR #1508, opened 2026-09-08
test(evidenceview): preserve timeline window chain positions

luckyPipewrench:
## What changed

Test that bounded receipt reads show the first timeline records, complete reads show the latest records, and both preserve the chain positions used to mark unverifiable evidence.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for receipt timeline-window behavior when chains contain verification breaks.
  * Verified that bounded, complete, exact-limit, and default-limit reads preserve receipt positions and return accurate verification results across different timeline sizes.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1511, opened 2026-09-08
test(support): retain large audit records in bundles

luckyPipewrench:
## What changed

Test that support bundles retain a large JSON audit record and the normal record that follows it, using output from the audit logger.

## Summary by CodeRabbit

- **Tests**
  - Added coverage to verify that bundled audit log tails preserve large JSON records without truncation or corruption.
  - Expanded validation for audit log output produced by the system, including records exceeding typical scanning limits.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1515, opened 2026-09-08
test: cover fail-closed paths in the signed-evidence chain

luckyPipewrench:
## What changed

Tests only, across the packages that produce and validate signed evidence. No production code changed.

The exported pre-sign sanitization boundary in `internal/receipt` had no direct tests, though ten production call sites stamp its results into a signed payload. These assert the property those callers depend on, which is that the returned value is clean under the predicate they supplied. The target shapes covered are a credential in userinfo, in a query value, split across two query values, in a path segment, in a fragment, in a host label, in a CONNECT authority, and in an MCP tool name. The documented redaction-off pass-through is pinned too, so it can't drift in either direction.

`validateRestartOpen` guards the seam where one run's receipt chain hands off to the next, and only its prior-tail-hash mismatch had a regression. A restart open that claims the genesis prefix, carries a genesis hash, or disagrees with its own receipt about where it sits now has to fail verification. Verifying any of those would present a gap as proven continuity.

`SignerKeyHex` is how reload decides whether a config change is a policy reload or a signer rotation, and it had no test. Two distinct keys have to report distinct values or a rotation is invisible. Retirement after a rotation has to brick the emitter so a stale caller can't append under the retired key, so the tests assert that refusal rather than the call's return value.

A coverage certificate's completeness status and reason are a coupled pair. Every status paired with another status's reason is rejected, along with a status outside the vocabulary and both directions of the receipt-count coupling. The violation has to be unsignable through `Sign`, not merely caught by the internal helper.

A relative posture-proof path is refused rather than resolved against the process working directory, which would leave the proof redirectable while still reading as attested. The two absent outcomes stay absent: an unconfigured path and a missing file must not read as tampering.

Every rejection case fails when the guard it covers is disabled, so none of them can pass vacuously.

## Summary by CodeRabbit

* **Tests**
  * Expanded validation coverage for completeness statuses, receipt counts, and signing safeguards.
  * Added checks for proof-path handling and availability reporting, including invalid, missing, and default paths.
  * Added lifecycle coverage for receipt emitters, key handling, retirement, abort behavior, and nil-safe operations.
  * Added sanitization tests covering clean values, detected secrets, URLs, authorities, and tool targets.
  * Added chain verification tests for restart-session integrity, trust-on-first-use, and untrusted signing keys.

### DIFF
--- a/internal/coveragecert/completeness_coupling_test.go
+++ b/internal/coveragecert/completeness_coupling_test.go
@@ -1,0 +1,147 @@
+// Copyright 2026 Josh Waldrep
+// SPDX-License-Identifier: Apache-2.0
+
+package coveragecert
+
+import (
+	"strings"
+	"testing"
+)
+
+// A coverage certificate's completeness_status and completeness_reason are a
+// coupled pair: the status says how much of the window is accounted for and
+// the reason says why. validateCompletenessCoupling is what stops a signed
+// certificate from claiming a pair that is not in the vocabulary, for example
+// LIMITED/chain_broken, which would understate a broken chain as a mere gap.
+//
+// This covers the whole invalid class rather than one instance: every status
+// paired with a reason belonging to a different status, an out-of-vocabulary
+// status, and both directions of the receipt_count coupling.
+func TestValidateCompletenessCoupling(t *testing.T) {
+	t.Parallel()
+
+	// Reasons grouped by the status that may legitimately carry them.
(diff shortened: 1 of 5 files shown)


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1516, opened 2026-09-08
fix(mcp): component-wise containment check, untrusted-input tests

luckyPipewrench:
## What changed

`isInsideDir` compared the relative path with a bare two-dot prefix test, so an ordinary child directory whose name begins with two dots read as outside the directory. Callers set the integrity report's suspicious flag from that result, and false means not suspicious, so a binary written to a path like `<workDir>/..cache/server` lost its warning. It now compares the parent reference as a whole path component, which is the form `internal/securefile` already uses for the same containment question.

That signal is advisory rather than enforcing. It's surfaced in the integrity report, it's deliberately zeroed when generating a manifest, and the proxy path passes an empty working directory so the comparison doesn't run there. The change restores a missing operator warning and moves toward reporting more rather than less.

The rest is coverage for input a verifier doesn't control. Message-signature extraction has several routes to not-found, and each one downgrades a signed message to unsigned instead of rejecting it, so these pin that no corrupted shape reports as verified and that the downgrade is reported as unsigned rather than passed over. The verifier's own nonce check can't be reached through our signing path, which refuses to emit a bad nonce, so it's covered from a message built as a struct instead.

Manifest parsing gained a rejection matrix. A null or absent entries field is rejected rather than read as a valid manifest that pins nothing. Duplicate JSON keys stay rejected because Go keeps the last occurrence while other readers keep the first, so one file could pin one set of hashes for this verifier and show a different set to anyone auditing it.

Every rejection case fails when the guard it covers is disabled, and the containment case fails against the previous prefix comparison, so none of them can pass vacuously.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved directory-boundary checks for similarly named paths, nested locations, and symbolic links.
  - Strengthened manifest validation by rejecting malformed, unsupported, duplicated, missing, or oversized data.
  - Improved handling of invalid or corrupted message signatures and envelopes so they are not treated as trusted.
  - Added validation for malformed security nonces while preserving existing metadata when adding signatures.

- **Tests**
  - Expanded coverage for path containment, manifest parsing, and untrusted message scenarios.

### DIFF
--- a/internal/mcp/integrity/containment_test.go
+++ b/internal/mcp/integrity/containment_test.go
@@ -1,0 +1,146 @@
+// Copyright 2026 Josh Waldrep
+// SPDX-License-Identifier: Apache-2.0
+
+package integrity
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// isInsideDir answers one containment question: is this resolved binary inside
+// the agent's own working directory? Callers set VerifyResult.Suspicious from
+// the answer, and false means "not suspicious", so every false is a warning
+// the operator does not see. The direction matters more than usual here.
+func TestIsInsideDir(t *testing.T) {
+	t.Parallel()
+
+	root := t.TempDir()
+	workDir := filepath.Join(root, "work")
+	if err := os.MkdirAll(filepath.Join(workDir, "bin"), 0o750); err != nil {
+		t.Fatalf("MkdirAll bin: %v", err)
+	}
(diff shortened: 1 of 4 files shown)


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1520, opened 2026-09-08
fix(config): bound best-effort expiry and per-agent authorization

luckyPipewrench:
## What changed

A configuration-sourced `sandbox.best_effort_expiry` may now lie at most 30 days after the time the config is validated. A later timestamp refuses the config, so one edit can't authorize the advisory network override for a year. Command-line durations are unchanged.

A per-agent `agents.<name>.sandbox` block that sets `best_effort: true` must carry its own `best_effort_reason` and `best_effort_expiry`, checked with the same rules as the top-level block, and it never inherits them from the top level. The profile's authorization travels with it into the merged config, a profile that turns the override off drops the top-level authorization, and the check runs before the enterprise license gate so an unlicensed load can't hide a malformed profile. An invalid config fails startup or reload with the reason. A failed reload keeps the previous config running.

## Summary by CodeRabbit

* **New Features**
  * Per-agent sandbox settings can override top-level controls and add filesystem paths.
  * Best-effort sandbox access can be authorized per agent with a required reason and expiration.
* **Bug Fixes**
  * Invalid sandbox authorizations are rejected consistently, including configurations that would otherwise be ignored.
  * Expiration values must be future RFC3339 timestamps within 30 days.
  * Changes to per-agent authorization details are detected during configuration reloads.
* **Documentation**
  * Updated sandbox configuration guidance for per-agent overrides and expiration limits.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1537, opened 2026-09-11
fix(mcp): judge what an agent-card change introduced before blocking

luckyPipewrench:
## What changed

A2A agent-card drift now evaluates what a card change introduced instead of blocking on the bare fact that the card changed. The baseline keeps two views of each card: a structural digest over its endpoint, auth, capability, mode and skill-identity fields, and the descriptive free text. Any structural change still blocks and holds the old baseline until an operator resets it. A descriptive change that introduces a new cue class, using the same discriminator MCP tool drift already uses, blocks and names the classes. A descriptive change that introduces nothing is adopted as the new baseline, stays clean, and is written to the audit log on every transport that scans cards, so an automatic trust update is never silent. A new skill counts as structural and blocks. Fields the card hash never covered, such as the provider block and documentation URL, stay outside drift detection exactly as before.

Skills with duplicate or empty IDs now sort by their remaining fields, so a reordered malformed card no longer reads as a structural change.

New tests pin two invariants that had no coverage: the MCP HTTP reverse listener reaches upstream only after a required receipt records, and not when the emitter is absent or the recorder fails, for both JSON and SSE responses; and a redirect-hop request-policy block on the fetch and forward paths advertises a receipt that actually recorded, and advertises nothing when recording failed. The redirect path already met that invariant, so no production behavior changed there.

## Summary by CodeRabbit

- **New Features**
  - A2A Agent Card scanning now distinguishes descriptive updates from structural changes.
  - Benign description-only updates are adopted as the new baseline without blocking responses.
  - Adopted updates are recorded in logs and audit events.

- **Bug Fixes**
  - Structural or endpoint changes, and descriptions introducing security-related cues, continue to block safely while preserving the previous baseline.
  - Unparseable cards fail closed, while version-only changes are ignored.
  - Redirect policy blocks now report receipt status accurately.
  - Receipt requirements are consistently enforced for standard and streaming responses.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1539, opened 2026-09-11
fix(scanner): resolve ordinary document-sharing links out of the box

luckyPipewrench:
## What changed

The path-entropy scanner blocked ordinary Google Docs, Sheets, Slides, Forms and Drive links, because those routes carry an opaque service-issued file ID that reads as a high-entropy path segment. An operator sent any of them could not open it, and the remedy closest to hand is disabling path entropy altogether: it is the only control that catches an unrecognizable payload in a URL path.

Five document-sharing routes now ship in the default `fetch_proxy.monitoring.path_entropy_exclusions`. Each binds one exact host to one literal path prefix and suppresses only the path-entropy gate; subdomain entropy, query entropy, DLP, the core credential floor and SSRF still run on the same request. A different route on the same host, the same route shape on another host, and a lookalike host all remain blocked. What a shipped entry encodes is the vendor's published route shape, never the identifier format, which Google documents as opaque.

`ApplyDefaults` now materializes the list, so a YAML-backed deployment inherits the routes rather than only a no-config run. An explicitly empty list remains a deliberate opt-out.

`pipelock doctor` no longer asks an operator to own, renew or remove a route Pipelock ships, since none of those is an action they can take. Cross-field warnings still fire on shipped entries, including the one that says a host-wide subdomain exclusion gives up subdomain-entropy coverage.

The canonical policy hash moves. That is deliberate: an operator's detection posture changed and the reported policy identity is what says so.

## Summary by CodeRabbit

* **New Features**
  * Added five default path-entropy exemptions for Google Docs, Sheets, Slides, Forms, and Drive sharing routes.
  * Empty configurations disable these defaults; custom lists replace them.
  * Other security checks remain active for exempted routes.

* **Bug Fixes**
  * Diagnostics now distinguish shipped defaults from operator-defined exclusions.
  * Unrelated, malformed, or lookalike routes remain blocked.

* **Documentation**
  * Documented defaults, customization behavior, and YAML examples.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -150,7 +150,7 @@ fetch_proxy:
 | `monitoring.subdomain_entropy_exclusions` | `files.pythonhosted.org`, `pypi.org`, `objects.githubusercontent.com` | Domains excluded from subdomain and path entropy checks; override to replace defaults, or set an empty list to disable exclusions entirely (query entropy still checked) |
 | `monitoring.scan_nested_urls` | `true` (nil) | Evaluate URL-shaped query parameter values as destinations |
 | `monitoring.query_entropy_exclusions` | `[]` | Host-wide query-string entropy exclusions for hosts whose query values are broadly opaque by contract |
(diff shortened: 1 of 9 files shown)


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1553, opened 2026-09-12
refactor(dashboard): compose authorizers through one registry

luckyPipewrench:
## What changed

Each dashboard authenticator (static and raw tokens, mapped client certificate, OIDC) now registers itself and one compose step installs the result, so adding a method no longer edits one inline block in the serve path. Decision outcomes, audit attribution and reported failure modes are unchanged for every token, mTLS and OIDC combination, and a table-driven test pins those outcomes as they were before the change.

The workbench template drift guard now also requires every conductor flag marked required to appear in the emitted command, instead of only checking that emitted flags exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard authorization consistency across static tokens, client certificates, and OIDC authentication.
  * Ensured client-certificate authentication is enforced consistently for metadata, raw access, permissions, audit details, and failed-auth responses.
  * Requests are denied by default when no dashboard authorizers are configured.

* **Tests**
  * Expanded coverage for combined authentication scenarios.
  * Strengthened conductor template validation for empty templates, undefined flags, and missing required rollback settings.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1554, opened 2026-09-13
fix(setup): report bundle-load problems from every integration hook

luckyPipewrench:
## What changed

- The proxy package resolves its own version from build info, so a source build reports the module version in startup logs and `/health` instead of a dev placeholder even when the CLI root never ran. A release stamp still wins.
- Cursor setup now reports bundle-load errors and warnings on stderr the way Claude setup already did, and both hooks share one reporter.
- A bundle rule that sets `exempt_domains` warns even when the rule is filtered out, and the rules guide says the field is ignored and exemptions belong in local config.

## Summary by CodeRabbit

* **New Features**
  * Bundle rules now warn when `pattern.exempt_domains` is configured but ignored.
  * Configure exemptions in the local Pipelock configuration instead.
  * Claude and Cursor hook setup reports bundle-loading warnings and errors to stderr while preserving JSON output.
  * Proxy builds now report consistent version information.

* **Bug Fixes**
  * Warnings are emitted even when affected rules are disabled or filtered.
  * Multiple ignored exemptions in a bundle are summarized in a single warning.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1581, opened 2026-09-15
feat(conformance): let containment fixtures supply nft chain text

luckyPipewrench:
## What changed

- A containment conformance probe fixture may now supply `nft_chain_text` with `agent_uid` and `proxy_uid` as an alternative to pre-baked `drop_counter_reads`, and that text runs through the same recognizer `contain verify` uses in production, which is now one shared function instead of a copy. Supplying both inputs is rejected as ambiguous, chain text without both UIDs is rejected, and the loaders reject unknown fields and a second JSON document.
- Two new must-fail fixtures cover outcomes the artifact couldn't express before: `agent-accept-before-drop` (a bare agent-UID accept ahead of the managed catch-all drop, a structural hole) and `dial-completed-then-failed` (curl failed after its connect completed). Expect files gain an optional `detail_contains` so failures with the same status are told apart. The two existing fixtures are byte-identical and still pass; the compatibility statement is in the testdata README.

## Summary by CodeRabbit

- **New Features**
  - Containment conformance checks now support literal firewall chain output alongside precomputed counter values.
  - Added validation for mutually exclusive inputs, required identifiers, unknown fields, trailing content, and valid ports.
  - Added richer probe expectations, including detail matching and fixture descriptions.

- **Bug Fixes**
  - Malformed or ambiguous containment evidence now fails closed instead of producing misleading results.
  - Added detection for traffic accepted before the managed drop rule and canaries that completed dialing before failing.

- **Documentation**
  - Updated fixture guidance, schema rules, and compatibility notes.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1582, opened 2026-09-15
feat(scanapi): accumulate scans across requests by session_id

luckyPipewrench:
## What changed

- The Scan API's `context.session_id` now does what the docs reserved it for: requests that share a session ID and the same bearer token accumulate content in a Scan-API-owned cross-request fragment buffer built from the existing `cross_request_detection.fragment_reassembly` settings, so a secret split across `dlp`, `prompt_injection` or `tool_call` requests is caught on the request that completes it. The response echoes the session ID and a completing finding lists the earlier scan IDs that contributed. The `url` kind doesn't participate, and requests without a session ID behave exactly as before.
- Failure direction: a completed match denies in block mode and escalates a clean result to warn in warn mode; capacity exhaustion and stream ownership conflicts deny regardless of action, and the denial names what recovers capacity. A malformed session ID is rejected with 400. Session identity folds a digest of the bearer token so two callers can't share a session, and the rate limiter now keys on that digest too instead of the raw token. Session state is local to one Scan API instance, and the docs say a multi-instance deployment must pin a caller's session to one instance.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in cross-request fragment reassembly for Scan API requests sharing a session.
  - Detects secrets split across DLP, prompt-injection, and tool-call requests.
  - Supports configurable session limits, caller isolation, and fail-closed behavior when capacity is exceeded.
  - Responses include session IDs and contributing scan IDs for cross-request findings.

- **Bug Fixes**
  - Invalid session IDs now return a clear `400 invalid_session_id` response.

- **Documentation**
  - Updated Scan API documentation with cross-request detection configuration and behavior.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1583, opened 2026-09-15
feat(playground): record requested and provider-reported model

luckyPipewrench:
## What changed

- The playground's signed launch manifest now records `requested_model`, the model string configured for the run, and the collector's signed witness records `provider_model`, the identifier the provider echoed on its first response. Both are optional and informational: the verifier never gates on them, artifacts without them verify unchanged, and an artifact carrying them needs a verifier from this version or later, the same additive pattern as the delegation and image-digest fields.
- The provider value is untrusted input. One shared helper bounds it to 256 printable ASCII bytes at the producer, before it crosses the agent subprocess event stream, and again before signing; a value that fails is recorded as empty with a warning and never fails the run or reaches visitor narration.

## Summary by CodeRabbit

- **New Features**
  - Playground runs now record configured and provider-reported model identifiers as informational provenance.
  - Model details are included in signed launch and evidence artifacts.
  - Provider identifiers are validated and bounded before recording.
  - Provider metadata is captured without appearing as visitor-facing narration.
  - Model-backed runs require the necessary provider connection settings.

- **Documentation**
  - Added guidance on model provenance, signature verification, and compatibility with legacy artifacts.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1584, opened 2026-09-15
feat(mcp): gate newly visible tool names under listener drift

luckyPipewrench:
## What changed

- MCP tool-drift detection trusted a tool name on first sighting, so an upstream could evade drift by adding a scan-clean new tool instead of changing an approved one. A new `mcp_tool_scanning.new_tool_action` knob (`warn` or `block`, default `warn`) now governs a name absent from an already established baseline: in block mode the definition is withheld from the baseline and reported with a `new-tool` cue on every later tools/list until the existing signed listener drift reset re-baselines; in warn mode it's admitted as before with a non-blocking observation that names the cue. The first valid inventory establishes the baseline whether or not it carries tools, through one atomic step that also resolves competing first responses on a shared baseline, and the capacity refusal still takes precedence.
- The knob is threaded through every transport's tool-scan config copy, validated at load and reload, and set explicitly in the shipped presets to match each preset's drift action. The canonical policy hash golden moves because the field is part of policy, the same way earlier policy fields moved it.

## Summary by CodeRabbit

* **New Features**
  * Added configurable handling for newly detected MCP tools: warn and admit them, or block them.
  * Blocked tools remain outside the trusted baseline and continue triggering drift alerts until the baseline is reset.
  * The initial tool inventory always establishes the baseline, regardless of this setting.
  * Presets now provide warning or blocking defaults.
  * Invalid settings are rejected, and weakening blocking settings during reload generates a warning.

* **Documentation**
  * Added guidance for new-tool admission, alerts, and baseline reset behavior.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1597, opened 2026-09-16
fix(license): make the trial slot table the sole eligibility authority

luckyPipewrench:
## What changed

- The trial slot table is the only authority for one active trial per email. The entitlement pre-count on the webhook path is gone, and the terminal sync writer that rewrote a slot's expiry is deleted.
- A slot's expiry is write-once. The same owner can refresh other columns, but an earlier period end can no longer move the slot into the past and free it while that owner's token is still valid.
- Trial issuance and a pending refund are mutually exclusive inside the transaction that claims the slot, so a refund recorded by one instance cannot race a mint on another. Startup lists slot rows whose expiry disagrees with the owning entitlement, and rows it cannot verify, without repairing them.

Instances running different versions of this service must not serve webhooks at the same time, because an older instance does not read the new guard. The reference deployment already runs one replica with the recreate strategy.

## Summary by CodeRabbit

- **New Features**
  - Trial eligibility now uses atomic slot checks for reliable concurrent orders and refunds.
  - Refunded or revoked trials retain their slot until the original claim expiry.
  - Trial claim expiry cannot be extended by repeat claims.
  - Added safeguards for refunds received before trial issuance completes.
  - Added detection and warning logs for inconsistent, unverifiable, or orphaned trial records requiring manual reconciliation.

- **Bug Fixes**
  - Trial fulfillment no longer depends on potentially unreliable entitlement counts.
  - Duplicate-trial denials are retried when audit recording fails.

- **Documentation**
  - Updated the trial fulfillment audit matrix.


luckyPipewrench:
@coderabbitai review

luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1600, opened 2026-09-16
fix(config): bound the expiry fields whose exception is temporary

luckyPipewrench:
## What changed

Seven configuration fields accepted any future expiry date, so a far-future value quietly turned a temporary exception into a permanent one, and the configuration guide demonstrated exactly that four times. Each field is now classified by whether the condition it excepts is expected to end.

- The six temporary ones carry a maximum horizon chosen per field from what that exception is for, enforced when the config loads and when it reloads. The refusal names the field, the configured date, the maximum, and what to do instead.
- A trusted upstream stays uncapped, because its condition does not expire and a cap there would manufacture re-approval work and push operators back toward far-future dates.

Configs that set one of the bounded fields beyond its horizon no longer load, and a reload carrying one keeps the previous policy active and reports the error. The documentation examples now use realistic dates.

## Summary by CodeRabbit

* **New Features**
  * Temporary expiry dates are now required for selected entropy exclusions, request-body exceptions, SigV4 credential routes, and unscannable response passthroughs.
  * Expiry windows are limited to 30–180 days, depending on configuration type.
  * Configuration reloads reject expired or overlong temporary authorizations.
  * Trusted upstream expiry dates remain uncapped, though expired dates still fail validation.

* **Documentation**
  * Updated configuration and tuning guidance with temporary expiry examples, limits, and renewal recommendations.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1615, opened 2026-09-18
docs(config): state the real default rule confidence

luckyPipewrench:
## What changed

The configuration reference said `rules.min_confidence` defaults to an empty value meaning every rule runs. It resolves to `medium`, both as the shipped default and through the normalizer that rewrites an omitted or empty value, so rules below medium confidence were being skipped while the documentation told an operator the opposite about their own detection coverage.

The table now states the real default and says that an omitted or empty value resolves to it.

## Summary by CodeRabbit

- **Documentation**
  - Updated configuration guidance to clarify that `rules.min_confidence` defaults to `medium`.
  - Clarified that omitted or empty values use the medium confidence level.


luckyPipewrench:
Folding this one-line correction into #1602, which is already open against the same docs area. Opening it separately was the wrong call: a single docs fix does not need its own review cycle.

PR #1624, opened 2026-09-19
fix(contain): keep the exported containment CA current and verifiable

luckyPipewrench:
## What changed

`contain install` skipped exporting the interception CA, so `/etc/pipelock/ca.pem` went stale after a rotation. The staleness was invisible because two Pipelock CAs can share the subject `O=Pipelock CN=Pipelock CA`: every check that compared subject names reported agreement while the exported file verified nothing the proxy presents. Install now refreshes the export on every run and compares certificate material rather than the subject name.

`NODE_EXTRA_CA_CERTS` pointed at that single-CA export while its six sibling CA variables pointed at the combined bundle, so Node inside a contained agent could trust a CA that signs nothing. It now uses the same bundle as every sibling, and the launch-contract check covers that variable, which it previously did not.

Adds verify probe 19, `pipelock_ca_export_current`. It is 19 rather than 18 because `contain run` already publishes 18 for `requested_tool_registered`, and a consumer keyed on the number could not otherwise tell the two results apart. That invariant was carried only by a comment that did not hold it, so a test now fails when the two number sets overlap.

Failure direction. Every check in the export step runs after the file is written, and the step reported no change on failure, so rollback never ran and a failed install could replace a good export with one it could not verify. `ca-refresh` wrote the export with no backup at all, so a failed bundle rebuild left `ca.pem` on the new CA while `combined-ca.pem` stayed on the old one, and clients follow the bundle. Both paths now restore the previous export.

The comparison is against the CA in the contain-managed keystore, not a live handshake: a proxy running since before a rotation still holds the previous CA in memory until it restarts.

## Summary by CodeRabbit

* **New Features**
  * Added verification that the exported Pipelock CA matches the certificate currently used by the proxy.
  * CA exports now refresh during every installation and detect certificate rotations.
  * Runtime clients, including Node.js, now consistently use the combined CA bundle.
  * Added checks for stale or invalid CA exports.

* **Bug Fixes**
  * CA files are now preserved or restored when updates or verification fail.

* **Documentation**
  * Documented CA trust management, refresh procedures, verification checks, and rotation behavior.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1627, opened 2026-09-19
fix(conductor): restore verified policy before serving

luckyPipewrench:
## What changed

Restore and re-verify cached Conductor policy before accepting traffic, refusing startup when the binary can't enforce it while preserving configured expiry and grace behavior.

## Summary by CodeRabbit

- **New Features**
  - Restores and verifies the last active policy configuration during server startup.
  - Applies stale-policy decisions immediately before requests are admitted, including after recovery.

- **Bug Fixes**
  - Detects cached configuration files that differ from their signed policy contents.
  - Prevents invalid, unauthorized, or unreadable cached policies from being restored.
  - Preserves the last known-good policy when recovery fails.
  - Ensures restored policies respect version, entitlement, signature, audience, and expiration requirements.


luckyPipewrench:
@coderabbitai review

Celest-Blue (approved):

PR #1639, opened 2026-09-21
fix: bound MCP and CI subprocess cleanup

luckyPipewrench:
## What changed

Protect active approval resolvers during MCP child cleanup and bound inherited output waits. Incomplete resolver output fails closed. Resolver completion terminates its original process group; it doesn't assign detached descendants to individual resolvers.

Bound each complete CI attempt and its output capture. Linux attempts terminate and reap their own detached descendants before returning. Python test timeouts also clean up detached log readers. Validate deadline arguments before arithmetic and stop retries after incomplete capture or cleanup.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of subprocesses and descendant processes during cancellation and concurrent operations.
  - Prevented commands from hanging indefinitely when inherited output streams remain open.
  - Improved reliability for deferred approval operations across success, failure, and cancellation scenarios.
  - Improved isolation so one operation’s cleanup does not terminate another operation’s processes.
  - Added bounded-time safeguards for background process cleanup.
  - Improved handling of blocked, allowed, and warned requests.

- **Chores**
  - Increased CI test time limits and strengthened cleanup, supervision, and retry handling for detached test processes.


luckyPipewrench:
@coderabbitai resume

luckyPipewrench:
Checked the review against `88ad7b8b7`. The Windows finding doesn't reproduce: `internal/filesentry/lineage_test.go` already has `//go:build linux`. With `GOOS=windows GOARCH=amd64`, `go list` puts it in `IgnoredGoFiles`, and `go test -c ./internal/filesentry` succeeds.

The reader-ownership candidates don't reproduce either. Controlled checks sent SIGTERM and SIGINT while the wrapper was blocked in each capture-reader wait, and between each reader's launch and PID assignment. All eight cases returned the expected 143 or 130 and collected the affected readers before fixture cleanup. During startup the traps record cancellation until ownership is assigned. During a reader wait the traps exit through `cleanup_on_exit` before the post-wait ownership assignments execute.

The earlier helper-path and capture-startup findings are fixed in this commit, with failing controls preserved. The earlier Python test-timeout finding is also fixed and covered by the passing wrapper tests. All 61 hosted checks passed on this head. The review recheck reports no verified findings but remains partial because its judge couldn't settle the reader-ownership candidate; these manual dispositions don't relabel that automated result.

Celest-Blue (approved):

